PR #12, opened 2020-03-21
add trace exporter example

mayurkale22:


mayurkale22:
@qingling128 Thanks for the reviews!

PR #23, opened 2020-04-02
chore: add trace example

mayurkale22:
Background: The Node.js samples that we are using in https://cloud.google.com/trace/docs/setup/nodejs-ot#auto-instrumentation_for_web_applications need to be hosted on GitHub. Also, this example contains region tags around the import content and the custom span content.


samples/trace/index.js
```diff
@@ -0,0 +1,75 @@
...
+// Enable OpenTelemetry exporters to export traces to Google Cloud Trace.
+// Exporters use Application Default Credentials (ADCs) to authenticate.
+// See https://developers.google.com/identity/protocols/application-default-credentials
+// for more details.
+// Expects ADCs to be provided through the environment as ${GOOGLE_APPLICATION_CREDENTIALS}
+// A Stackdriver workspace is required and provided through the environment as ${GOOGLE_PROJECT_ID}
+const projectId = process.env.GOOGLE_PROJECT_ID;
+
+// GOOGLE_APPLICATION_CREDENTIALS are expected by a dependency of this code
+// Not this code itself. Checking for existence here but not retaining (as not needed)
```
rghetia, line 38:
```suggestion
// and not this code itself. Checking for existence here but not retaining (as not needed)
```

samples/trace/index.js
```diff
@@ -0,0 +1,75 @@
...
+// [START opentelemetry_trace_import]
+const opentelemetry = require('@opentelemetry/api');
+const {NodeTracerProvider} = require('@opentelemetry/node');
+const {SimpleSpanProcessor} = require('@opentelemetry/tracing');
+const {
+  StackdriverTraceExporter,
+} = require('@google-cloud/opentelemetry-cloud-trace-exporter');
+// [END opentelemetry_trace_import]
+
+// [START setup_exporter]
```
rghetia, line 28: In Go examples the requirement was that these labels have same prefix. So the above should be changed to
`opentelemetry_trace_setup_expoter`

mayurkale22: @willgoldby could you please confirm?

samples/trace/index.js
```diff
@@ -0,0 +1,75 @@
...
+// Enable OpenTelemetry exporters to export traces to Google Cloud Trace.
+// Exporters use Application Default Credentials (ADCs) to authenticate.
+// See https://developers.google.com/identity/protocols/application-default-credentials
+// for more details.
+// Expects ADCs to be provided through the environment as ${GOOGLE_APPLICATION_CREDENTIALS}
+// A Stackdriver workspace is required and provided through the environment as ${GOOGLE_PROJECT_ID}
```
kintel, line 34: Perhaps remove references to "Stackdriver"?

samples/trace/index.js
```diff
@@ -0,0 +1,75 @@
...
+
+// [START opentelemetry_trace_samples]
+'use strict';
+
+// [START opentelemetry_trace_import]
+const opentelemetry = require('@opentelemetry/api');
+const {NodeTracerProvider} = require('@opentelemetry/node');
+const {SimpleSpanProcessor} = require('@opentelemetry/tracing');
+const {
+  StackdriverTraceExporter,
```
kintel, line 24: Since this is a new library, does it make sense to export an object referring to a historic product name?

samples/trace/index.js
```diff
@@ -0,0 +1,75 @@
...
+const span = tracer.startSpan('foo');
+
+// Set attributes to the span.
+span.setAttribute('key', 'value');
+
+// Annotate our span to capture metadata about our operation
+span.addEvent('invoking work');
+
+// Be sure to end the span.
+span.end();
```
kintel, line 70: To make this example a bit easier to follow, could we add some synthetic work here? A simple timeout would do I think.

samples/trace/index.js
mayurkale22: Good point, let me check what other languages are doing.

samples/trace/index.js
willgoldby: We did do that for the Go samples, but I'm not sure if it's a general rule. I know in other places it hasn't been required, so it looks like each repository has their own standards.

samples/trace/index.js
kintel: Worst case, we could keep it for backward compatibility with the old exporter and re-export a more suitable name, like CloudTraceExporter. ..and yes, we should sync this across languages.

mayurkale22: Changed to `TraceExporter` in samples and actual code (please review #24).

samples/trace/index.js
mayurkale22: Added empty for loop to simulate some random work work, PTAL

samples/trace/index.js
mayurkale22: done

kintel:
PS! Looks like `@opentelemetry/node` is missing from package.json.

mayurkale22:
> PS! Looks like `@opentelemetry/node` is missing from package.json.

Good catch, done.

PR #31, opened 2020-04-07
adding README.md to the trace

shivkanya9146:


samples/trace/README.md
```diff
@@ -0,0 +1,37 @@
+# Overview
+
+This example shows how to use [@opentelemetry/node](https://github.com/open-telemetry/opentelemetry-js/tree/master/packages/opentelemetry-node) to instrument a simple Node.js application - e.g. a batch job - and export spans either to [Google Cloud Trace](https://cloud.google.com/trace/).
+
+## Installation
+
+sh
```
mayurkale22, line 7: Please use triple backticks (```) before and after the code block.

See: https://help.github.com/en/github/writing-on-github/creating-and-highlighting-code-blocks

samples/trace/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Authenticate
+
+If you are running in a GCP environment, the exporter will automatically authenticate as the service account of your environment. Please make sure that it has permission to access Google Cloud Trace.
+
+If you are not running in a GCP environment you will need to create a service account and save the service account key json in the root of this example named `service_account_key.json`. For more information, visit <https://cloud.google.com/docs/authentication/>.
+
+## Run the Application
+
+sh
```
mayurkale22, line 20: same here.

samples/trace/README.md
```diff
@@ -0,0 +1,37 @@
+# Overview
+
+This example shows how to use [@opentelemetry/node](https://github.com/open-telemetry/opentelemetry-js/tree/master/packages/opentelemetry-node) to instrument a simple Node.js application - e.g. a batch job - and export spans either to [Google Cloud Trace](https://cloud.google.com/trace/).
```
kintel, line 3: Remove "either".
Also, it's generally good style to break long lines (same in multiple places)

samples/trace/README.md
```diff
@@ -0,0 +1,37 @@
...
+## Installation
+
+sh
+$ # from this directory
+$ npm install
+
+
+## Authenticate
+
+If you are running in a GCP environment, the exporter will automatically authenticate as the service account of your environment. Please make sure that it has permission to access Google Cloud Trace.
```
kintel, line 14: How would the user know it has permission to access Cloud Trace?

mayurkale22:
```suggestion
If you are running in a GCP environment, the exporter will automatically authenticate as the service account of your environment. However, you do need to ensure that your Google Cloud platform has the [Cloud Trace API access scope](https://developers.google.com/identity/protocols/googlescopes?&_ga=2.91528602.1456884696.1586202297-1073302523.1586202297#cloudtracev2) enabled.
```

@kintel what do you think about this?

kintel: If running inside GCP, I don't think Oauth2 scopes are applicable.

Perhaps we should just assume that this is running on GCP, as it's a lot less common to run tracing integrations outside of a GCP deployment, and we could save client-side instrumentation for a more advanced example.

mayurkale22: Are you suggesting to remove this (Authenticate) section?

kintel: Yeah, but with the caveat that it's not obvious to me how to run a stand-alone program like this in a GCP environment, so we may want to change it into being a service.

kintel: Anyway, this all exceeds the scope of this PR. Perhaps just kill the Authenticate section and open a new issue on making sure this example works out of the box.

PR #38, opened 2020-04-16
Adding-image in README.md file

shivkanya9146:
/cc @mayurkale22


samples/trace/README.md
```diff
@@ -20,7 +20,8 @@ $ npm start
 
 https://console.cloud.google.com/traces/list?project=your-project-id
 
-> TODO
+<img width="1584" alt="Trace_Waterfall_View" src="https://user-images.githubusercontent.com/62445341/79489186-e9361f80-7fe8-11ea-92af-2596ef8fed50.png">
```
mayurkale22, line 23: This is nice, but I would suggest to keep img in images folder inside `samples/trace/` and use something like this: `.png?raw=true"/>`

Issue #79, opened 2020-05-29
Monitoring: Metrics don't appear under any resource type


davidwitten:
*Is your feature request related to a problem? Please describe.*

I was exporting my metrics to Google Cloud. When I previously used OpenCensus and the Stackdriver Exporter, I was able to see my metrics when I selected a resource type.
![Screenshot 2020-05-29 at 6 02 42 PM](https://user-images.githubusercontent.com/3642085/83309205-a619bf80-a1d6-11ea-8d4a-3f95d69730ab.png) 

However, now I can only see data when I specify the "Global" resource type. 
![Screenshot 2020-05-29 at 5 26 46 PM - Display 2](https://user-images.githubusercontent.com/3642085/83309251-c9dd0580-a1d6-11ea-9364-ab2c98ec62bc.png)

*Describe the solution you'd like*
Within the Stackdriver exporter [code](https://github.com/census-instrumentation/opencensus-node/blob/master/packages/opencensus-exporter-stackdriver/src/common-utils.ts), they specify the resource type.  

Currently, this repository sets the resource to "Global" with a comment in the code [here](https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/blob/master/packages/opentelemetry-cloud-monitoring-exporter/src/transform.ts#L112) to wait until the OT API exists.

It seems that the OpenTelemetry resource API [exists now](https://github.com/open-telemetry/opentelemetry-js/tree/master/packages/opentelemetry-resources), so I'll try to add this feature and make a PR with this change.

mayurkale22:
Exporters can get access to `Resource` field via `MetricRecord`. This has been added in SDK in https://github.com/open-telemetry/opentelemetry-js/pull/1049. This feature is released in [v0.8.1](https://github.com/open-telemetry/opentelemetry-js/releases/tag/v0.8.1). I think it is good time to incorpoorate `Resource` into Monitoring exporter.

mayurkale22:
I can assign this issue to you, if you're working on it.

davidwitten:
Yeah, I'd be happy to work on this issue.

davidwitten:
[...]

PR #85, opened 2020-06-03
feat: Add resource within Google Cloud Monitoring Exporter

davidwitten:
# Summary
This PR is regarding Issue #79. 

Within createTimeSeries(...), I call detectResources, within @OpenTelemetry/resources. This determines whether it's an AWS EC2 Instance, a Google Cloud VM Instance, or another resource type. I convert it to a format which Stackdriver can accept.

I looked at the OpenCensus StackDriver exporter to determine the correct resource labels: https://github.com/census-instrumentation/opencensus-node/blob/master/packages/opencensus-exporter-stackdriver/src/common-utils.ts

I also updated the versions of the dependencies to allow for resource checking. This required a minor change (altering the type of one parameter). 

# Test Plan

I tested it locally on my Google Cloud VM Instance, and it successfully displayed metrics under the resource type "GCE VM Instance". 
Screenshot 2020-06-03 at 3 40 31 PM

## Problem setting up testing
Within transform.tests, I planned on mocking the environment to mimic AWS, GCP, and other. I planned on using Sinon to mock detectResources(...), but I couldn't because it's a standalone function. I had to also disable the resource check in the test. 

Do you have any suggestions for setting up testing for this code change?

### DIFF
--- a/packages/opentelemetry-cloud-monitoring-exporter/src/transform.ts
+++ b/packages/opentelemetry-cloud-monitoring-exporter/src/transform.ts
@@ -250,7 +250,6 @@ function transformValue(
   } else if (valueType === OTValueType.DOUBLE) {
     return { doubleValue: value };
   }
-  // TODO: Add support for Distribution and Histogram
   throw Error(`unsupported value type: ${valueType}`);
 }
 
--- a/packages/opentelemetry-cloud-monitoring-exporter/src/transform.ts
+++ b/packages/opentelemetry-cloud-monitoring-exporter/src/transform.ts
@@ -21,7 +21,11 @@ import {
   Point as OTPoint,
 } from '@opentelemetry/metrics';
 import { ValueType as OTValueType } from '@opentelemetry/api';
-import { Resource, CLOUD_RESOURCE, HOST_RESOURCE } from '@opentelemetry/resources';
+import {
+  Resource,
+  CLOUD_RESOURCE,
+  HOST_RESOURCE,
+} from '@opentelemetry/resources';
 import {
   MetricDescriptor,
   MetricKind,
--- a/packages/opentelemetry-cloud-monitoring-exporter/src/transform.ts
+++ b/packages/opentelemetry-cloud-monitoring-exporter/src/transform.ts
@@ -123,21 +123,21 @@ export async function createTimeSeries(
 
 async function transformResource(
   projectId: string
-  ): Promise<{ type: string; labels: { [key: string]: string } }> {
+): Promise<{ type: string; labels: { [key: string]: string } }> {
   const resource: Resource = await detectResources();
-  const cloudProvider: string = `${resource.labels['cloud.provider']}`;
+  const cloudProvider = `${resource.labels['cloud.provider']}`;
   // These are the only supported resources
   if (cloudProvider === 'gcp') {
     return {
       type: 'gce_instance',
       labels: {
         instance_id: `${resource.labels['host.id']}`,
-        project_id: projectId, 
+        project_id: projectId,
         zone: `${resource.labels['cloud.zone']}`,
       },
     };
   } else if (cloudProvider === 'aws') {
-    return { 
+    return {
       type: 'aws_ec2_instance',
       labels: {
         instance_id: `${resource.labels['host.id']}`,
--- a/packages/opentelemetry-cloud-monitoring-exporter/src/transform.ts
+++ b/packages/opentelemetry-cloud-monitoring-exporter/src/transform.ts
@@ -39,6 +39,7 @@ import * as os from 'os';
 
 const OPENTELEMETRY_TASK = 'opentelemetry_task';
 const OPENTELEMETRY_TASK_DESCRIPTION = 'OpenTelemetry task identifier';
+const AWS_REGION_VALUE_PREFIX = 'aws:';
 const GCP_GCE_INSTANCE = 'gce_instance';
 const AWS_EC2_INSTANCE = 'aws_ec2_instance';
 export const OPENTELEMETRY_TASK_VALUE_DEFAULT = generateDefaultTaskValue();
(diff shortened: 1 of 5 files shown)


packages/opentelemetry-cloud-monitoring-exporter/src/transform.ts
```diff
@@ -119,6 +121,35 @@ export function createTimeSeries(
...
+): { type: string; labels: { [key: string]: string } } {
+  const cloudProvider = `${resource.labels['cloud.provider']}`;
+  // These are the only supported resources
+  if (cloudProvider === 'gcp') {
+    return {
+      type: 'gce_instance',
+      labels: {
+        instance_id: `${resource.labels['host.id']}`,
+        project_id: projectId,
+        zone: `${resource.labels['cloud.zone']}`,
```
mayurkale22, line 136: I would suggest to use constants defined in OT JS instead of direct values, to avoid breaking change in future. https://github.com/open-telemetry/opentelemetry-js/blob/abd191bef366de8926f14701bf8b83df03b60d22/packages/opentelemetry-resources/src/constants.ts#L34

packages/opentelemetry-cloud-monitoring-exporter/src/transform.ts
```diff
@@ -119,6 +121,35 @@ export function createTimeSeries(
...
+      },
+    };
+  } else if (cloudProvider === 'aws') {
+    return {
+      type: 'aws_ec2_instance',
+      labels: {
+        instance_id: `${resource.labels['host.id']}`,
```
mayurkale22, line 143: same here and other places in this file.

packages/opentelemetry-cloud-monitoring-exporter/src/transform.ts
```diff
@@ -119,6 +121,35 @@ export function createTimeSeries(
   };
 }
 
+function transformResource(
+  resource: Resource,
+  projectId: string
+): { type: string; labels: { [key: string]: string } } {
+  const cloudProvider = `${resource.labels['cloud.provider']}`;
+  // These are the only supported resources
+  if (cloudProvider === 'gcp') {
```
mayurkale22, line 130: Do you think we should define constants for cloudProvider in OT JS and use it here?

Something like:
```js
export const CLOUD_PROVIDER = {
  GCP: 'gcp',
  AWS: 'aws',
  ....
};
```

davidwitten: Yeah I can make a followup task with @opentelemetry/resources to add that. Those are the only two cloud providers defined in @opentelemetry/resources.

packages/opentelemetry-cloud-monitoring-exporter/src/transform.ts
```diff
@@ -119,6 +127,35 @@ export function createTimeSeries(
...
+): { type: string; labels: { [key: string]: string } } {
+  const cloudProvider = `${resource.labels[CLOUD_RESOURCE.PROVIDER]}`;
+  // These are the only supported resources
+  if (cloudProvider === 'gcp') {
+    return {
+      type: GCP_GCE_INSTANCE,
+      labels: {
+        instance_id: `${resource.labels[HOST_RESOURCE.ID]}`,
+        project_id: projectId,
+        zone: `${resource.labels[CLOUD_RESOURCE.ZONE]}`,
```
mayurkale22, line 142: As per my experience with OpenCensus Stackdriver exporter, Stackdriver backend will complain expected labels are missing (see: https://github.com/census-instrumentation/opencensus-node/issues/706), which means we need to make sure all labels has valid values and in worst case fallback on global type.

You can very well copy code from OpenCensus: https://github.com/census-instrumentation/opencensus-node/blob/master/packages/opencensus-exporter-stackdriver/src/common-utils.ts

mayurkale22: You can easily reproduce this: 

1. Set type to `gce_instance`
2. Assign valid `instance_id` & `project_id` and keep zone as blank.
3. Try to export data to backend.

packages/opentelemetry-cloud-monitoring-exporter/src/transform.ts
```diff
@@ -33,6 +39,9 @@ import * as os from 'os';
 
 const OPENTELEMETRY_TASK = 'opentelemetry_task';
 const OPENTELEMETRY_TASK_DESCRIPTION = 'OpenTelemetry task identifier';
+const AWS_REGION_VALUE_PREFIX = 'aws:';
+const GCP_GCE_INSTANCE = 'gce_instance';
+const AWS_EC2_INSTANCE = 'aws_ec2_instance';
```
JorrieB, line 44: Optional: Given that these strings represent the actual monitored resource types, I'm curious your thoughts on putting them in an enum for reuse throughout the monitoring codebase.

packages/opentelemetry-cloud-monitoring-exporter/test/transform.test.ts
```diff
@@ -144,7 +145,36 @@ describe('transform', () => {
   });
 
   describe('TimeSeries', () => {
-    it('should return a Google Cloud Monitoring Metric', () => {
+    const mockedAwsResource = {
```
JorrieB, line 148: Optional, nit: I would go with mockAwsResource or even just awsResource, unless convention is `mocked` - mock just reads better to me and is equivalently clear in its meaning, omitting mocked altogether seems fine to me as well since we're in a test file and the implication is that these variables are for testing and no meaning would be lost in its brevity.

packages/opentelemetry-cloud-monitoring-exporter/src/transform.ts
```diff
@@ -119,6 +128,59 @@ export function createTimeSeries(
   };
 }
 
+function transformResource(
+  resource: Resource,
+  projectId: string
+): { type: string; labels: { [key: string]: string } } {
```
JorrieB, line 134: Consider using [MonitoredResource](https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/blob/ef3c4ad3eef08ec2c9e9f380e494248f13c73593/packages/opentelemetry-cloud-monitoring-exporter/src/types.ts#L68) as the return type here, getTypeAndMappings, and transformMetric to help readability.

mayurkale22:
@JorrieB Could you please review and approve the PR.

packages/opentelemetry-cloud-monitoring-exporter/src/transform.ts
```diff
@@ -119,6 +129,57 @@ export function createTimeSeries(
...
+  projectId: string
+): MonitoredResource {
+  const templateResource = getTypeAndMappings(resource);
+  const type = templateResource.type;
+  const labels: { [key: string]: string } = { project_id: projectId };
+
+  for (const key of Object.keys(templateResource.labels)) {
+    if (!resource.labels[templateResource.labels[key]]) {
+      return { type: 'global', labels: { project_id: projectId } };
+    } else {
```
JorrieB, line 143: Optional: Because you return in the if block above, consider removing the `else`. Doing so will flatten your structure as well.

packages/opentelemetry-cloud-monitoring-exporter/src/transform.ts
```diff
@@ -119,6 +129,57 @@ export function createTimeSeries(
...
+  resource: Resource,
+  projectId: string
+): MonitoredResource {
+  const templateResource = getTypeAndMappings(resource);
+  const type = templateResource.type;
+  const labels: { [key: string]: string } = { project_id: projectId };
+
+  for (const key of Object.keys(templateResource.labels)) {
+    if (!resource.labels[templateResource.labels[key]]) {
+      return { type: 'global', labels: { project_id: projectId } };
```
JorrieB, line 142:
```suggestion
      return { type: 'global', labels };
```
You've already defined labels as { project_id: projectId}.

JorrieB: I see that I might be wrong with my suggestion, given that you mutate labels. I'm opting to leave this comment instead as a question: is it your intention with this function to return only {type:'global', labels: {project_id: projectId}} if any of the labels are missing? I assume yes, but it's probably worth a function comment describing what the caller can expect if their resource is missing a key:value pair.

packages/opentelemetry-cloud-monitoring-exporter/src/transform.ts
```diff
@@ -119,6 +129,57 @@ export function createTimeSeries(
...
+function transformResource(
+  resource: Resource,
+  projectId: string
+): MonitoredResource {
+  const templateResource = getTypeAndMappings(resource);
+  const type = templateResource.type;
+  const labels: { [key: string]: string } = { project_id: projectId };
+
+  for (const key of Object.keys(templateResource.labels)) {
+    if (!resource.labels[templateResource.labels[key]]) {
```
JorrieB, line 141: Consider refactoring `resource.labels[templateResource.labels[key]]` into a local variable - it's referenced multiple times and is hard to read.

packages/opentelemetry-cloud-monitoring-exporter/src/transform.ts
```diff
@@ -119,6 +129,57 @@ export function createTimeSeries(
...
+        labels[key] = `${resource.labels[templateResource.labels[key]]}`;
+      }
+    }
+  }
+  return { type, labels };
+}
+
+function getTypeAndMappings(resource: Resource): MonitoredResource {
+  const cloudProvider = `${resource.labels[CLOUD_RESOURCE.PROVIDER]}`;
+  // These are the only supported resources
```
JorrieB, line 161: Optional: Consider specifying further what you mean by "These". Further, I think that this comment may work better as a function comment; that way, at first glance the caller knows which resources are supported by the function.

packages/opentelemetry-cloud-monitoring-exporter/test/transform.test.ts
```diff
@@ -155,18 +185,92 @@ describe('transform', () => {
...
       assert.strictEqual(ts.metricKind, MetricKind.CUMULATIVE);
       assert.strictEqual(ts.valueType, ValueType.DOUBLE);
       assert.strictEqual(ts.points.length, 1);
       assert.deepStrictEqual(ts.points[0].value, { doubleValue: 10 });
     });
+    it('should detect an AWS instance', () => {
+      const meter = new MeterProvider({
+        resource: new Resource(mockAwsResource),
+      }).getMeter('test-meter');
+      const labels: Labels = { ['keyb']: 'value2', ['keya']: 'value1' };
```
JorrieB, line 210: Optional: Reorder these labels, you monster.

PR #88, opened 2020-06-08
fix: Stackdriver batch limit for TimeSeries

### LINKED ISSUE #87: Add Batch Limit for Stackdriver Exporter
Quote by @mluggy:

"
On high-velocity metrics, you get:
UnhandledPromiseRejectionWarning: Error: Field timeSeries had an invalid value: A maximum of 200 TimeSeries can be written in a single request.

Apparently Stackdriver backend has the limit that each request cannot batch more than 200 time series so exporter needs to do some processing. For example https://github.com/census-instrumentation/opencensus-java/blob/9b2ac2bdc9565517155a9efb1b8581d1a91329dc/exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/CreateTimeSeriesExporter.java#L76-L77.

Using 0.0.16.
"

**Previous issue in OpenCensus:**
https://github.com/census-instrumentation/opencensus-node/issues/634

**Previous implementation in OpenCensus:**
https://github.com/census-instrumentation/opencensus-node/pull/644

TigerHe7:
**Description:**
* Adds a batch limit of 200 to TimeSeries sent due to limit in Stackdriver backend.

**Smaller changes:**
* Adds helper function for partitioning batches
* Fixes misleading comment in `transform.ts`
* Fixes parameter mismatch error being throw by `metricDescriptors` spy

**Previous implementation in OpenCensus:**
https://github.com/census-instrumentation/opencensus-node/pull/644

Fixes #87

### DIFF
--- a/packages/opentelemetry-cloud-monitoring-exporter/src/monitoring.ts
+++ b/packages/opentelemetry-cloud-monitoring-exporter/src/monitoring.ts
@@ -25,6 +25,11 @@ import { GoogleAuth, JWT } from 'google-auth-library';
 import { google } from 'googleapis';
 import { transformMetricDescriptor, createTimeSeries } from './transform';
 import { TimeSeries } from './types';
+import { partitionList } from './utils';
+
+// Stackdriver Monitoring v3 only accepts up to 200 TimeSeries per
+// CreateTimeSeries call.
+const MAX_BATCH_EXPORT_SIZE = 200;
 
 const OT_USER_AGENT = {
   product: 'opentelemetry-js',
@@ -111,7 +116,13 @@ export class MetricExporter implements IMetricExporter {
         );
       }
     }
-    this._sendTimeSeries(timeSeries);
+
+    for (const batchedTimeSeries of partitionList(
+      timeSeries,
+      MAX_BATCH_EXPORT_SIZE
+    )) {
+      await this._sendTimeSeries(batchedTimeSeries);
+    }
     cb(ExportResult.SUCCESS);
   }
 
--- a/packages/opentelemetry-cloud-monitoring-exporter/src/transform.ts
+++ b/packages/opentelemetry-cloud-monitoring-exporter/src/transform.ts
@@ -99,7 +99,7 @@ function transformValueType(valueType: OTValueType): ValueType {
 }
 
 /**
- * Converts metric's timeseries to a list of TimeSeries, so that metric can be
+ * Converts metric's timeseries to a TimeSeries, so that metric can be
  * uploaded to StackDriver.
  */
 export function createTimeSeries(
--- a/packages/opentelemetry-cloud-monitoring-exporter/src/utils.ts
+++ b/packages/opentelemetry-cloud-monitoring-exporter/src/utils.ts
@@ -1,0 +1,26 @@
+// Copyright 2020 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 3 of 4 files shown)


mayurkale22:
Did you get a chance to verify these changes against real instance?

Alss, please fix the lint.

packages/opentelemetry-cloud-monitoring-exporter/src/monitoring.ts
```diff
@@ -111,7 +116,13 @@ export class MetricExporter implements IMetricExporter {
         );
       }
     }
-    this._sendTimeSeries(timeSeries);
+
+    for (const batchedTimeSeries of partitionList(
+      timeSeries,
+      MAX_BATCH_EXPORT_SIZE
+    )) {
+      await this._sendTimeSeries(batchedTimeSeries);
```
TigerHe7, line 124: Added an `await` here for the async code in `_sendTimeSeries` to be properly called and spied on in tests.

I was wondering if there was a preferred testing practice so we don't have to wait here, or if it's ok to wait so we can handle if an error is thrown?

packages/opentelemetry-cloud-monitoring-exporter/src/transform.ts
```diff
@@ -99,7 +99,7 @@ function transformValueType(valueType: OTValueType): ValueType {
 }
 
 /**
- * Converts metric's timeseries to a list of TimeSeries, so that metric can be
+ * Converts metric's timeseries to a TimeSeries, so that metric can be
```
TigerHe7, line 102: Updated the comment here since a single TimeSeries is being returned. I think the discrepancy was made when this code was copied over.

packages/opentelemetry-cloud-monitoring-exporter/test/monitoring.test.ts
```diff
@@ -69,7 +71,7 @@ describe('MetricExporter', () => {
 
       metricDescriptors = sinon.spy(
         /* tslint:disable no-any */
-        (request: any, callback: (err: Error | null) => void): any => {
+        (request: any, params: any, callback: (err: Error | null) => void): any => {
```
TigerHe7, line 74: See spied on code on line 187 in `packages/opentelemetry-cloud-monitoring-exporter/src/monitoring.ts`

TigerHe7:
@mayurkale22 Did not get to testing against a real instance but will look into that now

packages/opentelemetry-cloud-monitoring-exporter/src/utils.ts
```diff
@@ -0,0 +1,24 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { TimeSeries } from './types';
+
+/** Returns an array with arrays of the given size. */
+export function partitionList(list: TimeSeries[], chunkSize: number) {
```
nlehrer, line 18: It would be better to copy the array first, since this will mutate the original array passed into the function. E.g. 
` const listCopy === [...list]`;

TigerHe7: Thanks for the feedback, updated!

packages/opentelemetry-cloud-monitoring-exporter/src/utils.ts
```diff
@@ -0,0 +1,24 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { TimeSeries } from './types';
+
+/** Returns an array with arrays of the given size. */
```
nlehrer, line 17: I think this could be more clear, like maybe "Returns the given array partitioned into arrays of the given size".

TigerHe7: Updated to "Returns the minimum number of arrays of max size chunkSize, partitioned from the given array"

PR #96, opened 2020-06-13
chore: add initial package skeleton for cloud-trace-propagator 

shivkanya9146:

Fixes #<issue_number_goes_here>


kintel:
Does it make sense for this to be a separate package? This is only really useful when using Google Cloud Trace, so it feels like incorporating it into the Cloud Trace exporter would make sense.

Along these lines: We should probably align this across languages. To my knowledge, Javascript is the first, but the Python one is in the works.

Caveat: I don't know anything about our packaging strategy

mayurkale22:
> This is only really useful when using Google Cloud Trace ...

I certainly agree with it. But you can use other OpenTelemetry propagators (such as `B3Format`, `HttpTraceContext`) and export to Google Cloud Trace backend. 

I reckon having an individual module / package would give great set of flexible to the end users. One known example is `cloud-trace-nodejs` library, which [depends on](https://github.com/googleapis/cloud-trace-nodejs/blob/master/package.json#L96) similar sort of a propagator. Moreover, our exporter package would be lightweight w/o propagator. I don't know if that matters. Having said that, I don't mind either way. @kintel WDYT?

kintel:
@mayurkale22 OK, after having looked through those other packages, it makes sense to make this a separate package.
I would suggest changing the package name to `@google-cloud/opentelemetry-cloud-trace-propagator`, as "Cloud Trace" is the product name we're supporting.

shivkanya9146:
> @mayurkale22 OK, after having looked through those other packages, it makes sense to make this a separate package.
> I would suggest changing the package name to `@google-cloud/opentelemetry-cloud-trace-propagator`, as "Cloud Trace" is the product name we're supporting.

Done:)

PR #97, opened 2020-06-16
feat: use gRPC to write spans to cloud trace

### LINKED ISSUE #94: Use gRPC to write spans to cloudtrace
*Is your feature request related to a problem? Please describe.*
As noted by @mayurkale22

The current implementation of the js trace exporter sends data to cloudtrace using HTTP:
https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/blob/5c6d7733e54f69aa1412ef9d0a9f8a2b99f6a4c1/packages/opentelemetry-cloud-trace-exporter/src/trace.ts#L98-L121

This is inconsistent with the other exporter implementations which use gRPC: 
https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/ec6b73c1970b009e48adce4d59d6dbbc97ba736e/exporter/trace/trace.go#L67

*Describe the solution you'd like*
Re-implement this to send data to cloudtrace using gRPC.

*Additional context*
Relevant usage of gRPC in the exporter collector: 
https://github.com/open-telemetry/opentelemetry-js/blob/837322ff05a7a8118a78f83b4f58be799ee5ce80/packages/opentelemetry-exporter-collector/src/platform/node/CollectorExporter.ts#L100

reggiemcdonald:
**Description**
- Adds a `cloudtrace.v2.TraceService` client to `TraceExporter` to be used to write span data to cloud trace
- Updated tests to work with the new rpc client

**Testing** 
- Passes existing unit tests
- Ran the following opentelemetry example found [here](https://github.com/open-telemetry/opentelemetry-js/tree/master/examples/grpc) modified to use the exporter in this PR, and it produced expected results
- Tested on a live instance

**Relevant Info**
- Docs for RPC service can be found [here](https://cloud.google.com/trace/docs/reference/v2/rpc/google.devtools.cloudtrace.v2#google.devtools.cloudtrace.v2.TraceService.BatchWriteSpans)

Fixes #94

### DIFF
--- a/packages/opentelemetry-cloud-trace-exporter/src/trace.ts
+++ b/packages/opentelemetry-cloud-trace-exporter/src/trace.ts
@@ -102,7 +102,9 @@ export class TraceExporter implements SpanExporter {
 
     const metadata = new grpc.Metadata();
     metadata.add(OT_REQUEST_HEADER, '1');
-    const batchWriteSpans = promisify(this._traceServiceClient.BatchWriteSpans);
+    const batchWriteSpans = promisify(
+      this._traceServiceClient.BatchWriteSpans
+    ).bind(this._traceServiceClient);
     try {
       await batchWriteSpans(spans, metadata);
       const successMsg = 'batchWriteSpans successfully';
@@ -134,7 +136,7 @@ export class TraceExporter implements SpanExporter {
         oneofs: true,
       }
     );
-    /* tslint:disable-next-line:no-any */
+    /* eslint-disable-next-line @typescript-eslint/no-explicit-any */
     const { google }: any = grpc.loadPackageDefinition(packageDefinition);
     const traceService = google.devtools.cloudtrace.v2.TraceService;
     const sslCreds = grpc.credentials.createSsl();
--- a/packages/opentelemetry-cloud-trace-exporter/src/trace.ts
+++ b/packages/opentelemetry-cloud-trace-exporter/src/trace.ts
@@ -19,7 +19,7 @@ import { Logger } from '@opentelemetry/api';
 import * as protoloader from '@grpc/proto-loader';
 import * as protofiles from 'google-proto-files';
 import * as grpc from '@grpc/grpc-js';
-import { GoogleAuth } from 'google-auth-library';
+import { GoogleAuth, UserRefreshClient, JWT, Compute } from 'google-auth-library';
 import { TraceExporterOptions } from './external-types';
 import { getReadableSpanTransformer } from './transform';
 import { TraceService, NamedSpans } from './types';
@@ -70,7 +70,15 @@ export class TraceExporter implements SpanExporter {
     this._logger.debug('Google Cloud Trace export');
 
     if (!this._traceServiceClient) {
-      await this._init();
+      let creds;
+      try {
+        creds = await this._auth.getClient();
+      } catch (err) {
+        err.message = `authorize error: ${err.message}`;
+        this._logger.error(err.message);
+        return resultCallback(ExportResult.FAILED_NOT_RETRYABLE);
+      }
+      this._init(creds);
     }
 
     const namedSpans: NamedSpans = {
@@ -121,16 +129,15 @@ export class TraceExporter implements SpanExporter {
   /**
    * Initializes the cloudtrace rpc client
    */
-  private async _init(): Promise<void> {
+  private _init(creds: Compute | JWT | UserRefreshClient): void {
     this._logger.debug('Google Cloud Trace initializing rpc client');
     const pacakageDefinition = protoloader.loadSync(protofiles.getProtoPath('devtools', 'cloudtrace', 'v2', 'tracing.proto'), {
       includeDirs: [protofiles.getProtoPath('..')],
     });
     const { google }: any = grpc.loadPackageDefinition(pacakageDefinition);
     const traceService = google.devtools.cloudtrace.v2.TraceService;
-    const creds = await this._auth.getApplicationDefault();
     const sslCreds = grpc.credentials.createSsl();
-    const callCreds = grpc.credentials.createFromGoogleCredential(creds.credential);
+    const callCreds = grpc.credentials.createFromGoogleCredential(creds);
     this._traceServiceClient = 
       new traceService('cloudtrace.googleapis.com:443', grpc.credentials.combineChannelCredentials(sslCreds, callCreds));
   }
--- a/packages/opentelemetry-cloud-trace-exporter/src/trace.ts
+++ b/packages/opentelemetry-cloud-trace-exporter/src/trace.ts
@@ -18,13 +18,8 @@ import { ReadableSpan, SpanExporter } from '@opentelemetry/tracing';
 import { Logger } from '@opentelemetry/api';
 import * as protoloader from '@grpc/proto-loader';
 import * as protofiles from 'google-proto-files';
-import * as grpc from '@grpc/grpc-js';
-import { 
-  GoogleAuth,
-  UserRefreshClient,
-  JWT,
-  Compute
-} from 'google-auth-library';
+import * as grpc from 'grpc';
+import { GoogleAuth, OAuth2Client } from 'google-auth-library';
 import { TraceExporterOptions } from './external-types';
 import { getReadableSpanTransformer } from './transform';
 import { TraceService, NamedSpans } from './types';
@@ -75,7 +70,7 @@ export class TraceExporter implements SpanExporter {
     this._logger.debug('Google Cloud Trace export');
 
     if (!this._traceServiceClient) {
-      let creds;
+      let creds: OAuth2Client;
       try {
         creds = await this._auth.getClient();
       } catch (err) {
@@ -134,7 +129,7 @@ export class TraceExporter implements SpanExporter {
   /**
    * Initializes the cloudtrace rpc client
    */
-  private _init(creds: Compute | JWT | UserRefreshClient): void {
+  private _init(creds: OAuth2Client): void {
     this._logger.debug('Google Cloud Trace initializing rpc client');
     const pacakageDefinition = protoloader.loadSync(
       protofiles.getProtoPath('devtools', 'cloudtrace', 'v2', 'tracing.proto'),
@@ -142,6 +137,7 @@ export class TraceExporter implements SpanExporter {
         includeDirs: [protofiles.getProtoPath('..')],
       }
     );
+    /* tslint:disable-next-line:no-any */
     const { google }: any = grpc.loadPackageDefinition(pacakageDefinition);
     const traceService = google.devtools.cloudtrace.v2.TraceService;
     const sslCreds = grpc.credentials.createSsl();
--- a/packages/opentelemetry-cloud-trace-exporter/src/trace.ts
+++ b/packages/opentelemetry-cloud-trace-exporter/src/trace.ts
@@ -20,6 +20,7 @@ import * as protoloader from '@grpc/proto-loader';
 import * as protofiles from 'google-proto-files';
 import * as grpc from 'grpc';
 import { GoogleAuth } from 'google-auth-library';
+import { promisify } from 'util';
 import { TraceExporterOptions } from './external-types';
 import { getReadableSpanTransformer } from './transform';
 import { TraceService, NamedSpans } from './types';
@@ -87,38 +88,31 @@ export class TraceExporter implements SpanExporter {
    * service.
    * @param spans
    */
-  private _batchWriteSpans(spans: NamedSpans): Promise<ExportResult> {
+  private async _batchWriteSpans(spans: NamedSpans): Promise<ExportResult> {
     this._logger.debug('Google Cloud Trace batch writing traces');
-    // Always resolve with the ExportResult code
-    return new Promise(async resolve => {
-      if (!this._traceServiceClient) {
-        try {
-          this._traceServiceClient = await this._getClient();
-        } catch (err) {
-          err.message = `authorize error: ${err.message}`;
-          this._logger.error(err.message);
-          return resolve(ExportResult.FAILED_NOT_RETRYABLE);
-        }
+    if (!this._traceServiceClient) {
+      try {
+        this._traceServiceClient = await this._getClient();
+      } catch (err) {
+        err.message = `authorize error: ${err.message}`;
+        this._logger.error(err.message);
+        return ExportResult.FAILED_NOT_RETRYABLE;
       }
+    }
 
-      const metadata = new grpc.Metadata();
-      metadata.add(OT_REQUEST_HEADER, '1');
-      this._traceServiceClient.BatchWriteSpans(
-        spans,
-        metadata,
-        (err: Error) => {
-          if (err) {
-            err.message = `batchWriteSpans error: ${err.message}`;
-            this._logger.error(err.message);
-            resolve(ExportResult.FAILED_RETRYABLE);
-          } else {
-            const successMsg = 'batchWriteSpans successfully';
-            this._logger.debug(successMsg);
-            resolve(ExportResult.SUCCESS);
-          }
-        }
-      );
-    });
+    const metadata = new grpc.Metadata();
+    metadata.add(OT_REQUEST_HEADER, '1');
+    const batchWriteSpans = promisify(this._traceServiceClient.BatchWriteSpans);
+    try {
+      await batchWriteSpans(spans, metadata);
+      const successMsg = 'batchWriteSpans successfully';
+      this._logger.debug(successMsg);
+      return ExportResult.SUCCESS;
+    } catch (err) {
+      err.message = `batchWriteSpans error: ${err.message}`;
+      this._logger.error(err.message);
+      return ExportResult.FAILED_RETRYABLE;
+    }
   }
 
   /**
--- a/packages/opentelemetry-cloud-trace-exporter/src/trace.ts
+++ b/packages/opentelemetry-cloud-trace-exporter/src/trace.ts
@@ -93,7 +93,7 @@ export class TraceExporter implements SpanExporter {
     try {
       this._traceServiceClient = await this._getClient();
     } catch (err) {
-      err.message = `authorize error: ${err.message}`;
+      err.message = `failed to create client: ${err.message}`;
       this._logger.error(err.message);
       return ExportResult.FAILED_NOT_RETRYABLE;
     }
--- a/packages/opentelemetry-cloud-trace-exporter/src/trace.ts
+++ b/packages/opentelemetry-cloud-trace-exporter/src/trace.ts
@@ -14,12 +14,17 @@
 
 import { ExportResult } from '@opentelemetry/base';
 import { NoopLogger } from '@opentelemetry/core';
-import { ReadableSpan, SpanExporter  } from '@opentelemetry/tracing';
+import { ReadableSpan, SpanExporter } from '@opentelemetry/tracing';
 import { Logger } from '@opentelemetry/api';
 import * as protoloader from '@grpc/proto-loader';
 import * as protofiles from 'google-proto-files';
 import * as grpc from '@grpc/grpc-js';
-import { GoogleAuth, UserRefreshClient, JWT, Compute } from 'google-auth-library';
+import { 
+  GoogleAuth, 
+  UserRefreshClient, 
+  JWT, 
+  Compute 
+} from 'google-auth-library';
 import { TraceExporterOptions } from './external-types';
 import { getReadableSpanTransformer } from './transform';
 import { TraceService, NamedSpans } from './types';
@@ -66,7 +71,7 @@ export class TraceExporter implements SpanExporter {
     if (!this._projectId) {
       return resultCallback(ExportResult.FAILED_NOT_RETRYABLE);
     }
-    
+
     this._logger.debug('Google Cloud Trace export');
 
     if (!this._traceServiceClient) {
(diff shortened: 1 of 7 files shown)


packages/opentelemetry-cloud-trace-exporter/src/trace.ts
```diff
@@ -95,49 +111,44 @@ export class TraceExporter implements SpanExporter {
...
-    }
+  private _init(creds: Compute | JWT | UserRefreshClient): void {
+    this._logger.debug('Google Cloud Trace initializing rpc client');
+    const pacakageDefinition = protoloader.loadSync(
+      protofiles.getProtoPath('devtools', 'cloudtrace', 'v2', 'tracing.proto'),
+      {
+        includeDirs: [protofiles.getProtoPath('..')],
+      }
+    );
+    const { google }: any = grpc.loadPackageDefinition(pacakageDefinition);
```
reggiemcdonald, line 145: The `any` typing here is causing the linter to fail. I'm not exactly sure how I can avoid using `any` since the proto-loader doesn't generate typings

reggiemcdonald:
@nlehrer

packages/opentelemetry-cloud-trace-exporter/src/trace.ts
```diff
@@ -70,17 +73,30 @@ export class TraceExporter implements SpanExporter {
     }
 
     this._logger.debug('Google Cloud Trace export');
-    const authorizedSpans = await this._authorize(
-      spans.map(getReadableSpanTransformer(this._projectId))
-    );
 
-    if (!authorizedSpans) {
+    if (!this._traceServiceClient) {
+      let creds;
```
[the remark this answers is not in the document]
reggiemcdonald: Yea good catch! I'll fix that up

packages/opentelemetry-cloud-trace-exporter/src/types.ts
```diff
@@ -133,8 +136,18 @@ export enum LinkType {
...
 
-export interface SpansWithCredentials {
+/**
+ * A protobuf boolean
+ */
+export interface BoolValue {
+  value: boolean;
+}
+
+export interface NamedSpans {
```
aabmass, line 146: I would expect there is a way to generate TS stubs from the protobuf to avoid having to manually make these types, and also the ones above like `Span`.

Looking through the available official grpc/proto stuff, its a bit unclear which library to use. It looks like there are 3 official implementations:

1. [grpc](https://github.com/grpc/grpc-node/blob/master/PACKAGE-COMPARISON.md)
    - deprecated
    - looks like this is the one in this PR
2. [@grpc/grpc-js](https://github.com/grpc/grpc-node/blob/master/PACKAGE-COMPARISON.md) - looks like the new version of #1, pure javascript
3. [grpc-web](https://github.com/grpc/grpc-web)
    - also pure javascript
    - can generate typescript
    - intended for the browser, would it work in node?

Then there are a few other OSS libraries that can generate TS
- https://github.com/improbable-eng/ts-protoc-gen
- https://github.com/protobufjs/protobuf.js#usage-with-typescript

Thoughts?

packages/opentelemetry-cloud-trace-exporter/src/trace.ts
```diff
@@ -70,17 +68,30 @@ export class TraceExporter implements SpanExporter {
...
+    if (!this._traceServiceClient) {
+      let creds: OAuth2Client;
+      try {
+        creds = await this._auth.getClient();
+      } catch (err) {
+        err.message = `authorize error: ${err.message}`;
+        this._logger.error(err.message);
+        return resultCallback(ExportResult.FAILED_NOT_RETRYABLE);
+      }
+      this._init(creds);
```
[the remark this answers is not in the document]
reggiemcdonald: Yes I think that that makes sense. What I was trying to go for here was distinguishing between a non-retryable authentication error and a retryable send error. But I still think its possible to account for that within the single function. I'll work on merging these

packages/opentelemetry-cloud-trace-exporter/src/trace.ts
```diff
@@ -95,49 +106,45 @@ export class TraceExporter implements SpanExporter {
...
-      err.message = `authorize error: ${err.message}`;
-      this._logger.error(err.message);
-      return null;
-    }
+  private _init(creds: OAuth2Client): void {
+    this._logger.debug('Google Cloud Trace initializing rpc client');
+    const pacakageDefinition = protoloader.loadSync(
+      protofiles.getProtoPath('devtools', 'cloudtrace', 'v2', 'tracing.proto'),
+      {
+        includeDirs: [protofiles.getProtoPath('..')],
```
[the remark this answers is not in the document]
reggiemcdonald: Ahh so I added a dependency here: 
https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/blob/94531e8f039f3c103c4d76234b7504be498c0e44/packages/opentelemetry-cloud-trace-exporter/package.json#L61
which provides googleapi proto files as an npm module which works nicely with the protoloader.

The convenience factor of this module is that it provides all the common types along with the specific APIs so you can load a full service definition out of the box

packages/opentelemetry-cloud-trace-exporter/src/transform.ts
```diff
@@ -26,9 +23,11 @@ import {
   Link,
   LinkType,
   Span,
+  Timestamp,
   TruncatableString,
 } from './types';
 import { VERSION } from './version';
+import { HrTime } from '@opentelemetry/api';
```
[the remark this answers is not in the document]
reggiemcdonald: Ahh yes, thats my mistake. Will fix!

packages/opentelemetry-cloud-trace-exporter/src/types.ts
reggiemcdonald: This is great feedback! 

1 (grpc) and 2 (@grpc/grpc-js) are functionally the same. In fact, you can simply change the imports and the code will work. The reason I chose the grpc package is that its compatible with node >=4, while @grpc/grpc-js requires node 8 or 12. I thought since the trace exporter's package.json says it requires node >=8 that the @grpc/grpc-js package had a narrower range then that.

grpc-web is specifically to make rpc calls from the browser. We cannot really use that because it requires a proxy server to convert request and response types between simple HTTP requests and rpc calls. 

Its definitely possible to use the TS stubs. In my experimentation I did actually implement the fully typed version using the protocol buffer compiler with the ts-protoc-gen plugin. I'd be happy to make this switch if people agree this is the way to go. One of the challenges I encountered with using the generated stubs is that it would require me to commit a fair number of files to the repo (~20 files: the proto files for google apis, the google common types they import, and their generated js and d.ts files).

I stuck with this implementation to start with because its consistent with what is being used in the OpenTelemetry [CollectorExporter](https://github.com/open-telemetry/opentelemetry-js/blob/bf2f6177eaaae416686c883c9efb076e862f29e9/packages/opentelemetry-exporter-collector/src/platform/node/CollectorExporter.ts#L79-L102) and this change is also quite minimal in comparison to the change required to use the generated stubs.

I haven't worked with protobufjs yet so I'll take a look!

packages/opentelemetry-cloud-trace-exporter/src/trace.ts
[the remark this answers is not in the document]
aabmass: Cool makes sense!

packages/opentelemetry-cloud-trace-exporter/src/types.ts
aabmass: Cool that all makes sense, sounds like this PR is a good compromise as is

packages/opentelemetry-cloud-trace-exporter/src/trace.ts
```diff
@@ -70,22 +68,14 @@ export class TraceExporter implements SpanExporter {
...
 
-    try {
-      await this._batchWriteSpans(authorizedSpans);
-      resultCallback(ExportResult.SUCCESS);
-    } catch (err) {
-      this._logger.error(`Google Cloud Trace failed to export ${err}`);
-      resultCallback(ExportResult.FAILED_RETRYABLE);
-    }
+    const result = await this._batchWriteSpans(namedSpans);
+    resultCallback(result);
```
reggiemcdonald, line 78: I dont think its necessary to wrap this in a try-catch because `_batchWriteSpans` now always resolves with an `ExportResult`

packages/opentelemetry-cloud-trace-exporter/test/exporter.test.ts
```diff
@@ -44,6 +44,21 @@ describe('Google Cloud Trace Exporter', () => {
     });
   });
 
+  describe('_getClient', () => {
```
nlehrer, line 47: Usually private methods are not unit tested, because they are implementation details, and any problems with them should be caught in tests of public methods. However, since `_getClient` is mocked in the 'export' describe block, the logic is not being tested at all, so I see why we might want this test. My suggestion would be instead of mocking `_getClient`, mock `google.devtools.cloudtrace.v2.TraceService` so that `new TraceService` will return the mock object that is currently returned from the mock `_getClient`. Then more of the real code will run when the 'should export spans' it block is run. You could also verify in that test that `new TraceService` is called with the correct two arguments (URL and credentials). Then I would remove this '_getClient' describe block and test. But let me know if you have any other thoughts/opinions on that.

reggiemcdonald: Yea this is a good suggestion. Will add

aabmass:
@reggiemcdonald I did a little more digging, `protobufjs` ([which is already a transitive dependency](https://github.com/grpc/grpc-node/blob/master/packages/proto-loader/package.json#L43) so no additional overhead to add) includes the scripts `pbts`/`pbjs` which can generate either functional code (so you don't need proto-loader) or just a `.d.ts` stub file. Some tradeoffs listed [here](https://github.com/protobufjs/protobuf.js#reflection-vs-static-code). As far as creating a bunch of files, I think you could actually do it without checking additional files in as a build step since `google-proto-files` gives you the protos like you mentioned.

Do you think it's doable?

reggiemcdonald:
> @reggiemcdonald I did a little more digging, `protobufjs` ([which is already a transitive dependency](https://github.com/grpc/grpc-node/blob/master/packages/proto-loader/package.json#L43) so no additional overhead to add) includes the scripts `pbts`/`pbjs` which can generate either functional code (so you don't need proto-loader) or just a `.d.ts` stub file. Some tradeoffs listed [here](https://github.com/protobufjs/protobuf.js#reflection-vs-static-code). As far as creating a bunch of files, I think you could actually do it without checking additional files in as a build step since `google-proto-files` gives you the protos like you mentioned.
> 
> Do you think it's doable?

Yep! I think that its possible and would be happy to make the change. Just playing around with protobufjs right now, it appears that [grpc cannot load protobufjs static files](https://github.com/grpc/grpc-node/issues/528), so I just need to figure out how to work with the protobufjs `rpcImpl`
@mayurkale22 what do you think?

kintel:
I'm not familiar with gRPC, so I cannot really comment on this, but it sounds like your investigation is going in the right direction.

reggiemcdonald:
I've implemented the protobufjs solution and tested it on a live instance. You can see the code [here](https://github.com/reggiemcdonald/opentelemetry-operations-js/tree/add-protobuf-impl/packages/opentelemetry-cloud-trace-exporter).

The reasons that I like this implementation are: 
- Types are automatically generated
- The protobufjs cli is an npm module and can be easily integrated into the build step without the need for external binaries

I left it off this branch for now in case this isn't the type of change that we want to make. Let me know what you think

aabmass:
I'm in favor but will defer to the folks who have been working in this repo longer :) This will also be useful for metrics

dancalvert-google:
Sounds reasonable to me as well, but I'm in a similar situation to Aaron as
far as having any authority to make this call.  Does anyone have concerns
with the approach?

mayurkale22:
At quick glance `proto-loader` approach looks good to me. That's what we did in OpenTelemetry collector exporter and in OpenCensus agent exporter.

aabmass:
@reggiemcdonald IIUC this won't work with web, is that right?

https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/issues/114

packages/opentelemetry-cloud-trace-exporter/test/exporter.test.ts
```diff
@@ -90,6 +92,38 @@ describe('Google Cloud Trace Exporter', () => {
...
+                cloudtrace: {
+                  v2: {},
+                },
+              },
+            },
+          };
+          // Replace the TraceService with a mock TraceService
+          def.google.devtools.cloudtrace.v2.TraceService = class MockTraceService
+            implements TraceService {
+            /* tslint:disable-next-line */
```
nlehrer, line 117: Is it possible to just disable the casing check? No worries if not, but just curious if there is something like disable-next-line:member-variable-must-be-lowerCamelCase

packages/opentelemetry-cloud-trace-exporter/test/exporter.test.ts
```diff
@@ -133,13 +167,53 @@ describe('Google Cloud Trace Exporter', () => {
...
-        batchWrite.getCall(0).args[0].resource.spans[0].displayName.value,
+        batchWrite.getCall(0).args[0].spans[0].displayName.value,
         'my-span'
       );
-
+      assert.strictEqual(traceServiceConstructor.getCalls().length, 1);
+      assert.strictEqual(
+        traceServiceConstructor.getCall(0).args[0],
+        'cloudtrace.googleapis.com:443'
+      );
```
nlehrer, line 177: Could we also test whether the second argument is what we would expect? Looking at the function body, I'm not sure how `grpc.credentials.createSsl()` or `grpc.credentials.createFromGoogleCredential(creds)` work. It seems like the latter uses the credentials returned from _getClient. So on line 92, maybe we could return some mock creds, and test if the output is what we would expect based on that mock. If the two `grpc.credentials` functions are non-deterministic somehow, then maybe it's more complicated, but let me know your thoughts.

packages/opentelemetry-cloud-trace-exporter/test/exporter.test.ts
reggiemcdonald: Yea good idea - after a bit of digging the following will work: 
```
/* tslint:disable-next-line:variable-name */
```

packages/opentelemetry-cloud-trace-exporter/test/exporter.test.ts
reggiemcdonald: In the grpc-node, the [credentials test](https://github.com/grpc/grpc-node/blob/ea36f751a1874ed119556fb6782c10bbaafa64b8/packages/grpc-native-core/test/credentials_test.js#L401-L415) uses an in-process rpc server to verify that google credentials worked. There doesn't appear to be much of an interface to validate internally that the credentials are good.

I was thinking that I could spy the ssl and google credentials to verify that they are being called with the appropriate parameters (none, and `auth.getClient()`)?

aabmass:
@reggiemcdonald is this ready for another review (going ahead with `proto-loader` approach), or were you planning more updates?

reggiemcdonald:
@aabmass should be ready for re-review!

packages/opentelemetry-cloud-trace-exporter/src/trace.ts
```diff
@@ -95,49 +85,64 @@ export class TraceExporter implements SpanExporter {
...
 
-    return new Promise((resolve, reject) => {
-      // @todo Consider to use gRPC call (BatchWriteSpansRequest) for sending
-      // data to backend :
-      // https://cloud.google.com/trace/docs/reference/v2/rpc/google.devtools.
-      // cloudtrace.v2#google.devtools.cloudtrace.v2.TraceService
-      TraceExporter._cloudTrace.projects.traces.batchWrite(
+      const metadata = new grpc.Metadata();
+      metadata.add('x-opentelemetry-outgoing-request', '1');
+      this._traceServiceClient.BatchWriteSpans(
```
aabmass, line 104: Looks to me like this is the only part of that needs to be converted into a promise, and the above code can be moved outside of the `new Promise(...)`? Maybe you can use [`util.promisifiy()`](https://nodejs.org/dist/latest-v8.x/docs/api/util.html#util_util_promisify_original)

 Too bad grpc doesn't support promises directly 🙁

packages/opentelemetry-cloud-trace-exporter/src/transform.ts
```diff
@@ -77,6 +75,13 @@ export function getReadableSpanTransformer(
   };
 }
 
+function transformTime(time: ot.HrTime): Timestamp {
```
aabmass, line 78: How does the `hrTimeToTimeStamp()` differ from this?

packages/opentelemetry-cloud-trace-exporter/src/types.ts
```diff
@@ -12,26 +12,31 @@
...
   links?: Links;
   status?: Status;
-  sameProcessAsParentSpan?: boolean;
+  sameProcessAsParentSpan?: BoolValue;
   childSpanCount?: number;
 }
 
+export interface Timestamp {
+  seconds?: number;
+  nanos?: number;
```
aabmass, line 37: Is there a place where these are undefined?

packages/opentelemetry-cloud-trace-exporter/src/trace.ts
```diff
@@ -95,49 +85,64 @@ export class TraceExporter implements SpanExporter {
...
-      this._logger.error(err.message);
-      return null;
-    }
+  private async _getClient(): Promise<TraceService> {
+    this._logger.debug('Google Cloud Trace authenticating');
+    const creds = await this._auth.getClient();
+    this._logger.debug(
+      'Google Cloud Trace got authentication. Initializaing rpc client'
+    );
+    const packageDefinition = protoloader.loadSync(
```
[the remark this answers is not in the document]
reggiemcdonald: Yep sounds good!

packages/opentelemetry-cloud-trace-exporter/src/transform.ts
reggiemcdonald: `hrTimeToTimeStamp()` returns a timestamp string, which is used on the REST endpoint. The grpc endpoint takes a [google.Protobuf.WellKnownTypes.Timestamp](https://cloud.google.com/trace/docs/reference/v2/rpc/google.devtools.cloudtrace.v2#span)

packages/opentelemetry-cloud-trace-exporter/src/types.ts
reggiemcdonald: I just made it consistent with the original definition of a timestamp where all fields are optional. Since we always set their value, I can make them mandatory here

packages/opentelemetry-cloud-trace-exporter/src/trace.ts
```diff
@@ -95,49 +85,64 @@ export class TraceExporter implements SpanExporter {
...
+    return new Promise(async resolve => {
+      if (!this._traceServiceClient) {
+        try {
+          this._traceServiceClient = await this._getClient();
+        } catch (err) {
+          err.message = `authorize error: ${err.message}`;
+          this._logger.error(err.message);
+          return resolve(ExportResult.FAILED_NOT_RETRYABLE);
+        }
+      }
```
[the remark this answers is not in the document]
reggiemcdonald: I need to resolve with `FAILED_NOT_RETRYABLE` when theres an authentication error. Do you have a suggestion on how to incorporate that into `_getClient()`?

packages/opentelemetry-cloud-trace-exporter/src/trace.ts
```diff
@@ -95,49 +85,64 @@ export class TraceExporter implements SpanExporter {
...
 
-    return new Promise((resolve, reject) => {
-      // @todo Consider to use gRPC call (BatchWriteSpansRequest) for sending
-      // data to backend :
-      // https://cloud.google.com/trace/docs/reference/v2/rpc/google.devtools.
-      // cloudtrace.v2#google.devtools.cloudtrace.v2.TraceService
-      TraceExporter._cloudTrace.projects.traces.batchWrite(
+      const metadata = new grpc.Metadata();
+      metadata.add('x-opentelemetry-outgoing-request', '1');
```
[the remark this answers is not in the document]
reggiemcdonald: I don't think that grpc converts them to HTTP headers. Right now, the gRPC-plugin can cause the same issue as described with the HTTP plugin in #117 

I've opened a PR for the [grpc/grpc-js plugins](https://github.com/open-telemetry/opentelemetry-js/pull/1353) to enable filtering of all requests that contain this metadata.

Do you foresee any issues caused by having this metadata?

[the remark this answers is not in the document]
aabmass: Awesome, just wanted to make sure that the metadata was being handled somewhere. Do you know if there is any update on the more general approach suggested here https://github.com/open-telemetry/opentelemetry-js/issues/332#issuecomment-660135431?

packages/opentelemetry-cloud-trace-exporter/src/trace.ts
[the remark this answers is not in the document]
aabmass: I just meant the outer `if (!this._traceServiceClient) { ... }`. Can this be moved into `_getClient()`?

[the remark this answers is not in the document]
aabmass: Also, we don't seem to check that the error is actually an auth error here. Do you need to prepend `authorize error: `?

packages/opentelemetry-cloud-trace-exporter/src/trace.ts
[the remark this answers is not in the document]
reggiemcdonald: It looks like the context issue is being [actively worked on](https://github.com/open-telemetry/opentelemetry-js/pull/1344)

packages/opentelemetry-cloud-trace-exporter/src/trace.ts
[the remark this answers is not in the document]
reggiemcdonald: Ahhhh okay sorry I wasn't following you there at first.

w.r.t the error message: I think the only thing in `getClient()` that can throw an error is the authentication, except for maybe a problem during npm install. But I've changed it to `failed to create client: ` which I think is most reflective of the issue

packages/opentelemetry-cloud-trace-exporter/src/trace.ts
```diff
@@ -95,49 +88,63 @@ export class TraceExporter implements SpanExporter {
...
-    });
+    const metadata = new grpc.Metadata();
+    metadata.add(OT_REQUEST_HEADER, '1');
+    const batchWriteSpans = promisify(
+      this._traceServiceClient.BatchWriteSpans
+    ).bind(this._traceServiceClient);
+    try {
+      await batchWriteSpans(spans, metadata);
+      const successMsg = 'batchWriteSpans successfully';
+      this._logger.debug(successMsg);
```
aabmass, line 109: nit no need for variable assignment here

packages/opentelemetry-cloud-trace-exporter/test/exporter.test.ts
```diff
@@ -75,21 +100,58 @@ describe('Google Cloud Trace Exporter', () => {
...
+                  v2: {},
+                },
+              },
+            },
+          };
+          // Replace the TraceService with a mock TraceService
+          def.google.devtools.cloudtrace.v2.TraceService = class MockTraceService
+            implements TraceService {
+            /* tslint:disable-next-line:variable-name */
+            BatchWriteSpans = batchWrite;
```
aabmass, line 146: is this why you need that `bind()` or is it needed for the actual call?

reggiemcdonald: Its needed for the actual call. Without it there is an undefined error within the grpc module

aabmass:
Awesome, thanks Reggie!

Issue #84, opened 2020-06-02
Linter: ESLint couldn't find a configuration file


davidwitten:
### What did you do?
I made changes to code and planned on running the linter before submitting a PR.

So, I ran `npm run lint`

### What did you expect to see?

This is an example when running in the opentelemetry-exporter-collector package:
```
> @opentelemetry/exporter-collector@0.7.0 lint /home/davidwitten_google_com/opentelemetry-js/packages/opentelemetry-exporter-collector
> gts check
```

### What did you see instead?
```
> @google-cloud/opentelemetry-cloud-monitoring-exporter@0.3.0 lint /home/davidwitten_google_com/opentelemetry-operations-js/packages/opentelemetry-cloud-monitoring-exporter
> gts check
version: 10
Oops! Something went wrong! :(
ESLint: 7.1.0
ESLint couldn't find a configuration file. To set up a configuration file for this project, please run:
    eslint --init
ESLint looked for configuration files in /home/davidwitten_google_com/opentelemetry-operations-js/packages/opentelemetry-cloud-monitoring-exporter/build/src and its ancestors. If it found none, it then looked in your home directory.
If you think you already have a configuration file or if you need more help, please stop by the ESLint chat room: https://eslint.org/chat
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! @google-cloud/opentelemetry-cloud-monitoring-exporter@0.3.0 lint: `gts check`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the @google-cloud/opentelemetry-cloud-monitoring-exporter@0.3.0 lint script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
npm ERR! A complete log of this run can be found in:
npm ERR!     /home/davidwitten_google_com/.npm/_logs/2020-06-02T21_00_59_109Z-debug.log
```

### Additional context
[...]

PR #107, opened 2020-06-30
fix: add local linting via gts [WIP]

TigerHe7:
Adding gts version 1.1.0 for local linting.

Ran `npx gts init` and updated `tsconfig.json` to point to `packages` directory.

Fixes #https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/issues/84


TigerHe7:
See @davidwitten's PR in the same issue

PR #109, opened 2020-07-02
fix: add local linting

jonahrosenblum:
**Description**

- Add linting through gts 1.1.0, the same version that CI runs. ESLint config copied and lightly modified from main opentelemetry-js repo.
- Updated one unit test to no longer use the deprecated .equals() function which threw a linter error. This is the only linter error the CI does not pick up on but I think it's worth including.

**Testing**

- Modified unit tests still work as intended

**Relevant Info**

- Addresses issue #84


packages/opentelemetry-cloud-monitoring-exporter/package.json
```diff
@@ -39,11 +39,11 @@
   "devDependencies": {
     "@types/mocha": "7.0.2",
     "@types/nock": "11.1.0",
-    "@types/node": "12.12.47",
     "@types/sinon": "9.0.4",
     "codecov": "3.7.0",
-    "gts": "2.0.2",
+    "gts": "1.1.0",
     "mocha": "7.2.0",
+    "nock": "12.0.3",
```
aabmass, line 46: Was this just missing altogether before? Also the one below has `^`, should they be consistent?

packages/opentelemetry-cloud-monitoring-exporter/package.json
```diff
@@ -39,11 +39,11 @@
   "devDependencies": {
     "@types/mocha": "7.0.2",
     "@types/nock": "11.1.0",
-    "@types/node": "12.12.47",
     "@types/sinon": "9.0.4",
     "codecov": "3.7.0",
-    "gts": "2.0.2",
+    "gts": "1.1.0",
```
aabmass, line 44: What's the reason to use this older version?

packages/opentelemetry-cloud-monitoring-exporter/package.json
jonahrosenblum: Before nock was imported from @types/nock. I could be misremembering but I believe this gave me an error when using node 8.

packages/opentelemetry-cloud-monitoring-exporter/package.json
jonahrosenblum: When the CI runs for linting these are the modules installed:
+ lerna@3.22.1
 
+ gts@1.1.0
 
+ typescript@3.9.5
 
I thought it would make sense for the local linting to use the same version of gts as the CI.

packages/opentelemetry-cloud-monitoring-exporter/package.json
jonahrosenblum: The reason for using this specific version is that it was working in opentelemetry-js/opentelemetry-resources which heavily uses nock successfully.

packages/opentelemetry-cloud-monitoring-exporter/package.json
aabmass: I see, looks like it was pinned in https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/pull/36 which links to some instructions for upgrading. If you got it working on 2.x, I don't see any reason to not upgrade?

jonahrosenblum: My original thinking was that going forward it might make sense to have the local version of gts tied to the one in CI in case anything comes up locally which would not be considered a "real issue" by the CI linting. Currently there exist no discrepancies so I agree there is no reason not to keep it at 2.0.2, I reverted it.

aabmass: I meant upgrade it in the CI config as well. If there are no issues with the new version, we can remove that 1.1.0 dep from the CI lint config so local and CI do the same thing, right?

.circleci/config.yml
```diff
@@ -37,7 +37,7 @@ jobs:
       - checkout
       - run:
           name: Install minimal modules globally
-          command: npm i -g lerna gts@1.1.0 typescript
+          command: npm i -g lerna gts@2.0.2 typescript
```
aabmass, line 40: I wonder why it was originally running `npm i -g`. This should work right?
```suggestion
          command: npm i
```
Then it will just use the same version as package.json when the next step runs `npm run lint`

jonahrosenblum: I believe it was attempting to install the minimum number of required packages to make running CI faster, but yes I can change it.

jonahrosenblum: It seems as though it installs from a different package.json, I believe the best option currently might be to just stay with gts 1.1.0.

aabmass:
@jonahrosenblum I dug into this a little bit, looks like the CI jobs are using the `node:12` docker image:
https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/blob/83c7966eead63a1b9a7f5d7da79edfef0fb0382f/.circleci/config.yml#L33-L35

I tried out this image locally and it runs as the root user. Looks like npm doesn't run postinstall scripts automatically as root user as a security concern (I think `--ignore-scripts` will also skip this, so you probably don't want that flag). One option is to manually run the bootstrap like the other CI jobs do.

I think CircleCI also suggests using their docker images (like [circleci/node:12](https://hub.docker.com/layers/circleci/node/12/images/sha256-2c0e12e221431d4b45fadaa5fe6e4449013a327b447af5997e127328c1567938?context=explore)) which are quite small and don't run as root user. It's up to you how you want to solve this :)

mayurkale22:
@dancalvert-google @kintel FYI. I am not sure who else to add to review this PR.

.circleci/config.yml
```diff
@@ -32,12 +32,12 @@ unit_tests: &unit_tests
...
   lint:
     docker:
-      - image: node:12
+      - image: circleci/node:12
     steps:
       - checkout
       - run:
           name: Install minimal modules globally
-          command: npm i -g lerna gts@1.1.0 typescript
+          command: npm install --ignore-scripts
```
reggiemcdonald, line 40: The `npm i -g lerna gts@1.1.0 typescript` is installing global executables to be used to run the lint. 

You can use `npm install` but you dont want to have the `--ignore-scripts` flag because that will only install the top level directory. The three packages wont be installed, and the `gts` dev dependency wont be available

Now that you've changed the container to not run as root, I think all you need here is `npm install`

aabmass:
@mayurkale22 if there was any reason to use the `node:*` docker images over the `circleci/node:*` ones? I think we can remove some of the setup code if we change them all over to not run as root

jonahrosenblum:
@aabmass @reggiemcdonald Let me know if this is ok, using just `npm install` almost worked but it required package-lock files to be present so I removed them from the .gitignore.

packages/opentelemetry-cloud-monitoring-exporter/package.json
```diff
@@ -1,9 +1,9 @@
 {
```
aabmass, line 1: Any reason to reformat to 4 spaces?

jonahrosenblum: This was a mistake, my text editing program acts out sometimes.

mayurkale22:
> @mayurkale22 if there was any reason to use the `node:*` docker images over the `circleci/node:*` ones? I think we can remove some of the setup code if we change them all over to not run as root

This is just a carry over from OpenCensus and OpenTelemetry, moreover, I am not 100% familiar with `circleci/node: *`. I am fine with changing it to `circleci/node:*`.

jonahrosenblum:
> > @mayurkale22 if there was any reason to use the `node:*` docker images over the `circleci/node:*` ones? I think we can remove some of the setup code if we change them all over to not run as root
> 
> This is just a carry over from OpenCensus and OpenTelemetry, moreover, I am not 100% familiar with `circleci/node: *`. I am fine with changing it to `circleci/node:*`.

Is this something I should address in this PR or should we open up a separate issue for this?

mayurkale22:
If you ask me, I would prefer in the separate PR.

PR #110, opened 2020-07-06
fix: cloud monitoring incompatibility with opentelemetry/metrics 0.9.0

### LINKED ISSUE #102: cloud monitoring exporter is not compatible with the latest release of @opentelemetry/metrics
@opentelemetry/metrics@0.9.0

13

### What did you do?
Starting with a new directory, copied and ran the sample code in the README [here](https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/blob/master/packages/opentelemetry-cloud-monitoring-exporter/README.md).

I installed two dependencies: 
- `@opentelemetry/metrics@0.9.0`
- `@google-cloud/opentelemetry-cloud-monitoring-exporter`

and added the appropriate ADC for google cloud.

Finally, I made a logger set to debug, and passed it as an option to the `MetricExporter`

The metrics should have been sent to cloud monitoring and visible within the console.

### What did you see instead?
An error message was thrown instead, and the metrics were not uploaded: 
```
Google Cloud Monitoring export
TypeError: Cannot read property 'map' of undefined
    at transformLabelDescriptor (/Users/reginaldmcdonald/Documents/github/metrics-example/node_modules/@google-cloud/opentelemetry-cloud-monitoring-exporter/build/src/transform.js:46:43)
    at Object.transformMetricDescriptor (/Users/reginaldmcdonald/Documents/github/metrics-example/node_modules/@google-cloud/opentelemetry-cloud-monitoring-exporter/build/src/transform.js:32:17)
    at MetricExporter._createMetricDescriptor (/Users/reginaldmcdonald/Documents/github/metrics-example/node_modules/@google-cloud/opentelemetry-cloud-monitoring-exporter/build/src/monitoring.js:118:35)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
    at async MetricExporter._registerMetricDescriptor (/Users/reginaldmcdonald/Documents/github/metrics-example/node_modules/@google-cloud/opentelemetry-cloud-monitoring-exporter/build/src/monitoring.js:99:30)
    at async MetricExporter.export (/Users/reginaldmcdonald/Documents/github/metrics-example/node_modules/@google-cloud/opentelemetry-cloud-monitoring-exporter/build/src/monitoring.js:72:34)
done
```

In typescript, the above example will produce a type conflict, and will not compile. 

If I swap out `@opentelemetry/metrics@0.9.0` with an older 0.6.1 release, then the code runs fine and the metric is visible from the cloud console. 

### Additional context
This is likely to do with a change to the opentelemetry spec where labelKeys have been removed [here](https://github.com/open-telemetry/opentelemetry-js/commit/c24d9923054fff431feb215c47c6c5647973dfe6#diff-0cdf7b503ad3cf97f7b836b0cf885210) and [here](https://github.com/open-telemetry/opentelemetry-js/commit/c24d9923054fff431feb215c47c6c5647973dfe6#diff-e3728f5692d938914d82d335cd641267)

We can see that in the metric descriptor transformer, it's assumed that this key still exists: 
https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/blob/3ecb2e240e66f5b8cfbfa65d6a62ada3329feda3/packages/opentelemetry-cloud-monitoring-exporter/src/transform.ts#L38-L52

which results in the above type error when the following code is executed: 
https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/blob/3ecb2e240e66f5b8cfbfa65d6a62ada3329feda3/packages/opentelemetry-cloud-monitoring-exporter/src/transform.ts#L65-L77

reggiemcdonald:
This PR upgrades the cloud monitoring exporter's dependency on @opentelemetry/metrics to the latest release (0.9.0).

### Changes
The biggest change to @opentelemetry/metrics was the [removal](https://github.com/open-telemetry/opentelemetry-js/pull/1126) of `labelKeys` from `MetricOptions`. In response, labels are now added to GCP metrics using the metric instead of the metric descriptor as per the [GCP Docs](https://cloud.google.com/monitoring/custom-metrics/creating-metrics#md-modify)

Functions were adjusted to handle the additional `MetricKind`s that were [added](https://github.com/open-telemetry/opentelemetry-js/pull/1145)

### Testing 
Added some additional unit tests; all tests pass. 
Tested metric export on a GCE VM running the latest version of node.

Fixes #102

### DIFF
--- a/packages/opentelemetry-cloud-monitoring-exporter/src/monitoring.ts
+++ b/packages/opentelemetry-cloud-monitoring-exporter/src/monitoring.ts
@@ -117,11 +117,21 @@ export class MetricExporter implements IMetricExporter {
       }
     }
 
+    let sendFailed: Error | undefined;
     for (const batchedTimeSeries of partitionList(
       timeSeries,
       MAX_BATCH_EXPORT_SIZE
     )) {
-      await this._sendTimeSeries(batchedTimeSeries);
+      try {
+        await this._sendTimeSeries(batchedTimeSeries);
+      } catch (err) {
+        err.message = `Send TimeSeries failed: ${err.message}`;
+        sendFailed = err;
+        this._logger.error(err.message);
+      }
+    }
+    if (sendFailed) {
+      return cb(ExportResult.FAILED_RETRYABLE);
     }
     cb(ExportResult.SUCCESS);
   }
--- a/packages/opentelemetry-cloud-monitoring-exporter/src/monitoring.ts
+++ b/packages/opentelemetry-cloud-monitoring-exporter/src/monitoring.ts
@@ -117,7 +117,7 @@ export class MetricExporter implements IMetricExporter {
       }
     }
 
-    let sendFailed: Error | undefined;
+    let sendFailed = false;
     for (const batchedTimeSeries of partitionList(
       timeSeries,
       MAX_BATCH_EXPORT_SIZE
@@ -125,9 +125,9 @@ export class MetricExporter implements IMetricExporter {
       try {
         await this._sendTimeSeries(batchedTimeSeries);
       } catch (err) {
-        err.message = `Send TimeSeries failed: ${err.message}`;
-        sendFailed = err;
-        this._logger.error(err.message);
+        const message = `Send TimeSeries failed: ${err.message}`;
+        sendFailed = true;
+        this._logger.error(message);
       }
     }
     if (sendFailed) {
--- a/packages/opentelemetry-cloud-monitoring-exporter/src/transform.ts
+++ b/packages/opentelemetry-cloud-monitoring-exporter/src/transform.ts
@@ -24,7 +24,6 @@ import { ValueType as OTValueType } from '@opentelemetry/api';
 import {
   MetricDescriptor,
   MetricKind,
-  LabelDescriptor,
   ValueType,
   TimeSeries,
   Point,
--- a/packages/opentelemetry-cloud-monitoring-exporter/src/transform.ts
+++ b/packages/opentelemetry-cloud-monitoring-exporter/src/transform.ts
@@ -68,11 +68,11 @@ function transformDisplayName(displayNamePrefix: string, name: string): string {
 function transformMetricKind(kind: OTMetricKind): MetricKind {
   switch (kind) {
     case OTMetricKind.COUNTER:
-    case OTMetricKind.SUM_OBSERVER:
-      return MetricKind.CUMULATIVE;
     case OTMetricKind.UP_DOWN_COUNTER:
+      return MetricKind.CUMULATIVE;
     // OTMetricKind.OBSERVER will be removed in opentelemetry-js #1146
     case OTMetricKind.OBSERVER:
+    case OTMetricKind.SUM_OBSERVER:
     case OTMetricKind.VALUE_OBSERVER:
     case OTMetricKind.UP_DOWN_SUM_OBSERVER:
       return MetricKind.GAUGE;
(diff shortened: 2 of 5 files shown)


packages/opentelemetry-cloud-monitoring-exporter/src/monitoring.ts
```diff
@@ -117,11 +117,21 @@ export class MetricExporter implements IMetricExporter {
       }
     }
 
+    let sendFailed: Error | undefined;
```
nlehrer, line 120: Could this just be a boolean? I think that might be more readable.

E.g.
```
let sendFailed = false;
...
catch (err) {
  sendFailed = true;
}
```

packages/opentelemetry-cloud-monitoring-exporter/src/monitoring.ts
```diff
@@ -117,11 +117,21 @@ export class MetricExporter implements IMetricExporter {
...
       timeSeries,
       MAX_BATCH_EXPORT_SIZE
     )) {
-      await this._sendTimeSeries(batchedTimeSeries);
+      try {
+        await this._sendTimeSeries(batchedTimeSeries);
+      } catch (err) {
+        err.message = `Send TimeSeries failed: ${err.message}`;
+        sendFailed = err;
+        this._logger.error(err.message);
```
nlehrer, line 130: I think it might be more readable and less confusing behavior to create a new variable with the modified message to be logged rather than mutating the original error. I think modifying the caught error wouldn't do much unless it is re-thrown. Not a big deal either way, though.

packages/opentelemetry-cloud-monitoring-exporter/test/transform.test.ts
```diff
@@ -168,5 +173,124 @@ describe('transform', () => {
...
+        metricDescriptor,
+        new Date().toISOString()
+      );
+
+      assert.deepStrictEqual(result.value, { int64Value: 50 });
+      assert(result.interval.endTime);
+      assert(!result.interval.startTime);
+    });
+
+    it('should throw an error when given a distriution value', () => {
```
nlehrer, line 234: typo: distriution -> distribution

packages/opentelemetry-cloud-monitoring-exporter/src/transform.ts
```diff
@@ -147,37 +142,58 @@ function transformPoint(
...
+  value: number | OTDistribution | OTHistogram
 ) {
+  if (isDistributionValue(value)) {
+    // TODO: Add support for Distribution
+    throw new Error('Distributions are not yet supported');
+  }
+  if (isHistogramValue(value)) {
+    // TODO: Add support for Histogram
+    throw new Error('Histograms are not yet supported');
+  }
```
nlehrer, line 177: This is inconsistent with the existing code, which uses the `valueType` rather than the value to distinguish the behavior. I think there are pros and cons to each; using the `valueType` isn't as type safe as using the `value`, but it is simpler. I would recommend using `valueType` for now and perhaps revisiting when distributions/histograms are implemented. Does that make sense to you? Having explicit cases for each with TODOs and specific errors still looks good to me.

reggiemcdonald: I agree that it would be a better solution to just switch on `valueType` if `valueType` could be either int, double, histogram, or distribution. From what I understand however, `valueType` can only ever be int or double (see [link to code](https://github.com/open-telemetry/opentelemetry-js/blob/904245f7678537d9f3ae1ce31bd7a8b9469cc7db/packages/opentelemetry-api/src/metrics/Metric.ts#L64-L67)). I believe that switching solely on `valueType` would cause unexpected behaviour when `value` is a `Distribution` or `Histogram`. For example when `valueType` is `INT` and `value` is a `Distribution`, the output of the function would have the following structure: 
```
{
  int64Value: {
    min: <int>,
    max: <int>,
    count: <int>,
    sum: <int>,
  }
}
```
which I suspect will cause a problem when the payload gets sent to the timeseries.create api.

Its also possible that I'm misunderstanding how valuetypes and values relate to one another.

nlehrer: I see, I misunderstood how that works. I was having trouble locating the ValueType enum, so thanks for sending that. 

So I think what you have here is good. What I would suggest is making a [user-defined type guard](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards) for the functions checking the types. E.g.:

```
function isDistributionValue(value: number | OTDistribution | OTHistogram): value is OTDistribution {
  return value.hasOwnProperty('min');
}
```

Then I think you could safely remove the `as number` type casts in 180 and 182 since it would be narrowed down to number.

That would also help later working with those distribution and histogram types.

reggiemcdonald:
@danherr

packages/opentelemetry-cloud-monitoring-exporter/src/transform.ts
reggiemcdonald: Great idea! Will do

Issue #101, opened 2020-06-24
Add distribution and histogram metric support


nlehrer:
Only non-distribution (bool, int64, double, etc.) values are supported by the exporter at the moment. See [this](https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/blob/a1f6264bec877bc00c66348d249ac076d8cf0207/packages/opentelemetry-cloud-monitoring-exporter/src/transform.ts#L177) line of code for the relevant TODO and where support should be added (perhaps among other places).

nlehrer:
Check if a distribution without buckets is rejected by the API.

coreyvan:
Are histograms and distribution metric kinds supported now? I noticed that a fix got merged in as part of #125 but there's still this method that sets the `MetricKind` to `UNSPECIFIED` (which causes GCP to reject the metric descriptor) here: https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/blob/60f90dccefe2f903b67c28a61cd78300fae4737a/packages/opentelemetry-cloud-monitoring-exporter/src/transform.ts#L72-L85

nlehrer:
It does seem from that code that it is still not supported but that it could be easy to complete the support. A new owner needs to be found for this issue since @TigerHe7 is no longer working on this project (neither am I).

PR #125, opened 2020-07-24
feat: add support for Distributions

TigerHe7:
Addresses: https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/issues/101

### Summary
Adding support for exporting OpenTelemetry Distribution and Histogram types as Cloud Monitoring API v3 Distributions

### Tests
* Tested manually with exporting to personal GCP project
* Updated unit tests

### DIFF
--- a/packages/opentelemetry-cloud-monitoring-exporter/src/transform.ts
+++ b/packages/opentelemetry-cloud-monitoring-exporter/src/transform.ts
@@ -237,23 +237,15 @@ function transformValue(
   value: number | OTDistribution | OTHistogram
 ) {
   if (isDistributionValue(value)) {
-    // minimum two buckets due to mandatory buckets params in Cloud Monitoring API v3
-    return {
-      distributionValue: {
-        count: value.count,
-        mean: value.sum / value.count,
-        // sumOfSquaredDeviation param not aggregated
-        bucketOptions: { explicitBuckets: { bounds: [value.min] } },
-        bucketCounts: [0, value.count],
-      },
-    };
+    throw Error('unsupported distribution value type');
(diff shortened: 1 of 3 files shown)


packages/opentelemetry-cloud-monitoring-exporter/src/transform.ts
```diff
@@ -237,12 +237,21 @@ function transformValue(
...
   if (isHistogramValue(value)) {
-    // TODO: Add support for Histogram
-    throw new Error('Histograms are not yet supported');
+    return {
+      distributionValue: {
+        // sumOfSquaredDeviation param not aggregated
+        count: value.count,
+        mean: value.sum / value.count,
+        bucketOptions: {
+          explicitBuckets: { bounds: value.buckets.boundaries },
```
aabmass, line 250: Just want to double check that the JS SDK and Cloud Monitoring are using the same histogram boundary format. IIUC Cloud Monitoring uses a format where `[1, 5, 10]` would give intervals `(-inf, 1), [1, 5), [5, 10), [10, inf)` https://cloud.google.com/monitoring/api/ref_v3/rpc/google.api#explicit

Is OT JS using the same format?

TigerHe7: Hi @aabmass, sorry for the delay. From my testing, it seemed to me like they were using the same histogram boundary format, but I found that some inline comments which seemed to suggest the opposite when I double-checked after your comment. Those inline comments were incorrect and I made a PR to fix them / add a test: https://github.com/open-telemetry/opentelemetry-js/pull/1427

So in conclusion, the JS SDK and Cloud Monitoring are using the same histogram boundary format :)

Issue #129, opened 2020-07-28
Add support for Histogram


TigerHe7:
**Change Requested:**
Add support for `OTMetricKind.ValueRecorder` from OpenTelemetry SDK.

**Current Behavior:**
Currently, trying to export metrics collected via `ValueRecorder` results in an error from Google Cloud Monitoring regarding the `MetricKind.UNSPECIFIED` being used.

aabmass:
The OTel metrics specification still hasn't decided what default aggregator will be used for ValueRecorder. It's looking like it will be "some histogram sketch" (e.g. DDSketch) which will produce fancy histograms. Since this will be changing, we should put this work off until the JS SDK is up to date.

jonnydgreen:
Looking at the spec feature freeze milestone, this is due by the end of September 2021: https://github.com/open-telemetry/opentelemetry-specification/milestone/14. If you need, happy to help you both out on an implementation when the default aggregator has been chosen for ValueRecorder and the JS SDK is up to date :)

aabmass:
We're always happy to receive PRs 😃  I think this is completely blocked until this work is done in the spec (September timeline) AND its implemented in the JS SDK (maybe later).

PR #130, opened 2020-07-28
feat: add support for OT ValueRecorder

TigerHe7:
Fixes: https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/issues/129

### Changes
* Adds support for exporting `ValueRecorder` distribution metrics
* Tested use of `ValueRecorder` manually on a personal GCP project
* Updated unit tests

### Feedback Request
* The `transformValueType` function needs the `metric.aggregator.toPoint().value` to disambiguate between `Distribution`/`Histogram` data types. Please let me know if you think this should be passed down to `transformValueType` in a different manner than shown.


packages/opentelemetry-cloud-monitoring-exporter/src/transform.ts
```diff
@@ -86,17 +87,18 @@ function transformMetricKind(kind: OTMetricKind): MetricKind {
...
-      // OTMetricKind.Measure was renamed to ValueRecorder in #1117
       return MetricKind.UNSPECIFIED;
   }
 }
 
 /** Transforms a OpenTelemetry ValueType to a StackDriver ValueType. */
-function transformValueType(valueType: OTValueType): ValueType {
-  if (valueType === OTValueType.DOUBLE) {
+function transformValueType(valueType: OTValueType, point: OTPoint): ValueType {
+  if (isDistributionValue(point.value) || isHistogramValue(point.value)) {
```
aabmass, line 99: I'm not familiar with the JS SDK, but looking through this code, `isDistribution()` will be true for `ValueRecorder`. Is that right?

I'm confused because `transformValue()` will then throw on L242.

TigerHe7: Yeah, that's correct. I've opened a separate PR to address the support for `Distribution` and `Histogram` type metrics to be exported here: https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/pull/125

When testing with those changes, I've gotten the `ValueRecorder` to correctly export.

nicolasgere:
Who is responsible of merging that?

aabmass:
This PR needs to be updated and it should take a minor change to send LastValue

aabmass:
The OTel metrics specification still hasn't decided what default aggregator will be used for ValueRecorder. It's looking like it will be "some histogram sketch" (e.g. DDSketch) which will produce fancy histograms. Since this will be changing, closing this PR.

PR #159, opened 2020-08-25
fix: adjust peer dep versions of OT packages

### LINKED ISSUE #156: 0.4.2 requires deprecated @opentelemetry/base@0.7.0 to be installed

0.10.2

v12

### What did you do?

Install 0.4.2 version with opentelemetry 0.10.2

Compile without issues

### What did you see instead?

Got module not found error for "@opentelemetry/base" in cloud-trace-exporter code

### Additional context

Deprecated peer dependency @opentelemetry/base is required (https://www.npmjs.com/package/@opentelemetry/base). It has been merged in @opentelemetry/core now. The deprecated package @opentelemetry/base@0.7.0 can be added in package.json but I am not sure if mixing @opentelemetry/base@0.7.0 and @opentelemetry/core@0.10.2 will be stable as both will contain different versions of same code. 

For now, I have kept exporter at 0.4.1 (thus having a whole separate copy of @opentelemetry at 0.7.0) and propagator at 0.4.2 (as that only works with peer dependencies due to the way opentelemetry context works)

The solution, I think, is to update the dependencies of cloud-trace-exporter to 0.8.0 or later when @opentelemetry/base and @opentelemetry/core have been merged and fix the require statements therein.

aabmass:

Fixes #156 

- Removed references to `@opentelemetry/base` which was merged into `@opentelemetry/core` for 0.8.0.
- Then, upgraded all min peer/dev dependencies to be atleast `0.8.0` if they weren't already.
- Capped peer dependency versions at current `0.10.x`; will test more releases as they come out and increase minor version.
- Upgraded sample app versions and tested. `npm list` does not output any errors about peer deps.
- Added CI step to test that sample app dependencies have no errors, like unsatisfied peer deps that we previously had.

### DIFF
--- a/packages/opentelemetry-cloud-trace-exporter/package.json
+++ b/packages/opentelemetry-cloud-trace-exporter/package.json
@@ -61,9 +61,9 @@
     "grpc": "^1.24.3"
   },
   "peerDependencies": {
-    "@opentelemetry/api": ">=0.8.0 <=0.10.x",
-    "@opentelemetry/core": ">=0.8.0 <=0.10.x",
-    "@opentelemetry/resources": ">=0.8.0 <=0.10.x",
-    "@opentelemetry/tracing": ">=0.8.0 <=0.10.x"
+    "@opentelemetry/api": "0.8.0 - 0.10.x",
+    "@opentelemetry/core": "0.8.0 - 0.10.x",
+    "@opentelemetry/resources": "0.8.0 - 0.10.x",
+    "@opentelemetry/tracing": "0.8.0 - 0.10.x"
   }
 }
--- a/packages/opentelemetry-cloud-trace-exporter/package.json
+++ b/packages/opentelemetry-cloud-trace-exporter/package.json
@@ -37,11 +37,9 @@
     "registry": "https://wombat-dressing-room.appspot.com"
   },
   "devDependencies": {
-    "@opentelemetry/api": "0.7.0",
-    "@opentelemetry/base": "0.7.0",
-    "@opentelemetry/core": "0.7.0",
-    "@opentelemetry/resources": "0.7.0",
-    "@opentelemetry/tracing": "0.7.0",
+    "@opentelemetry/api": "0.8.0",
+    "@opentelemetry/resources": "0.8.0",
+    "@opentelemetry/tracing": "0.8.0",
     "@types/mocha": "7.0.2",
     "@types/node": "12.12.51",
     "@types/sinon": "9.0.4",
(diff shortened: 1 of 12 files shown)


samples/trace/package.json
```diff
@@ -11,9 +11,9 @@
   },
   "repository": "GoogleCloudPlatform/opentelemetry-operations-js",
   "dependencies": {
-    "@google-cloud/opentelemetry-cloud-trace-exporter": "0.3.0",
-    "@opentelemetry/api": "0.7.0",
-    "@opentelemetry/tracing": "0.7.0",
-    "@opentelemetry/node": "0.7.0"
+    "@google-cloud/opentelemetry-cloud-trace-exporter": "file:../../packages/opentelemetry-cloud-trace-exporter",
```
mayurkale22, line 14: Is this intentional ?

samples/trace/README.md
```diff
@@ -5,6 +5,8 @@ This example shows how to use [@opentelemetry/node](https://github.com/open-tele
 ## Installation
 
 ```sh
+$ # from root of repo, build all packages
+$ npm install
```
mayurkale22, line 9: Why do we need to run this command on the root? I thought samples are always running against published version?

packages/opentelemetry-cloud-trace-exporter/package.json
```diff
@@ -37,11 +37,9 @@
     "registry": "https://wombat-dressing-room.appspot.com"
   },
   "devDependencies": {
-    "@opentelemetry/api": "0.7.0",
-    "@opentelemetry/base": "0.7.0",
-    "@opentelemetry/core": "0.7.0",
-    "@opentelemetry/resources": "0.7.0",
-    "@opentelemetry/tracing": "0.7.0",
+    "@opentelemetry/api": "0.8.0",
```
mayurkale22, line 40: The latest version is `0.10.2`

samples/trace/package.json
aabmass: Yes, so you can run the sample against the repo's version. I can change it back, but then it's hard to verify the sample works with the current commit.

samples/trace/README.md
aabmass: Since I changed the sample to use the repo version, it needs to be compiled. Not necessary if it runs against the published version

packages/opentelemetry-cloud-trace-exporter/package.json
aabmass: Should it be kept at the oldest working version or the newest working version?

mayurkale22: If you ask me, I would  directly put new version. I think this is fine till we are in Beta version.

samples/trace/README.md
mayurkale22: I see, I don't have a strong objection to this.

PR #175, opened 2020-09-29
feat: upgrade packages to opentelemetry version 0.11.0

### LINKED ISSUE #165: Add support for opentelemetry 0.11.0
In #159 the opentelemetry packages were required to be between 0.8 and 0.10, however a day later version 0.11 of the opentelemetry packages was released.

That version has a breaking change as `HttpTextPropagator` was renamed to `TextMapPropagator`. However, given that the opentelemtry packages and these package are still in beta (0.x versions), I'd suggest to keep them in sync.

One could consider having the same versioning scheme while they are both in beta. For example having version `0.11.0` of `@google-cloud/opentelemetry-cloud-trace-propagator` to have a peer dependency on `@opentelemetry/*` `0.11.x`. In that way it's easier to know whether the dependencies are compatible.

aabmass:

Fixes #165

----

It is no longer possible to support back to v0.8.0 because of a fair number of breaking changes. As suggested in #165, this upgrades all peer dependencies to `^0.11.0`. The next release will be a "major" release (v0.5.0 per semantic versioning < v1.0.0) because of these breaking changes.

Fixes for new opentelemetry version:

- `resource.labels -> resource.attributes`
- `OTPoint` is now generic `OTPoint<PointValueType>`
- `meter.collect()` needs to now be awaited in tests. I'm not sure if it was previously not async, but the tests fail if without waiting for the promise to resolve.
- `OTMetricKind.OBSERVER` was removed https://github.com/open-telemetry/opentelemetry-js/issues/1146
- **`SumObserver` appears to now be `CUMULATIVE`**
  * I don't think this requires any other changes on our side, but would be great to get a second opinion
  * Another effect of this is that we now get `startTime`
- `HttpTextPropagator -> TextMapPropagator`

### DIFF
--- a/packages/opentelemetry-cloud-monitoring-exporter/package-lock.json
+++ b/packages/opentelemetry-cloud-monitoring-exporter/package-lock.json
@@ -300,115 +300,50 @@
       "dev": true
     },
     "@opentelemetry/api": {
-      "version": "0.8.3",
-      "resolved": "https://registry.npmjs.org/@opentelemetry/api/-/api-0.8.3.tgz",
-      "integrity": "sha512-CRd8kKrMN1yisbw2zc2IiXM31zdS8mvQ4sivxPFgWS/eQEW9XuNxMVtSWv5k/tDcqbcLzUHcX6JW0SpVNA5h+g==",
+      "version": "0.11.0",
+      "resolved": "https://registry.npmjs.org/@opentelemetry/api/-/api-0.11.0.tgz",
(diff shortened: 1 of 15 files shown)


mayurkale22 (approved):
LGTM

> SumObserver appears to now be CUMULATIVE

This looks fine to me as sum never decreases, so makes sense to me report it as CUMULATIVE kind.

aabmass:
Not sure why codecov went down, even after adding another test

PR #189, opened 2020-10-28
chore: upgrade for opentelemtry js v0.12.0

### LINKED ISSUE #188: Upgrade to opentelemetry 0.12.0
A new verdion of opentelemetry was released, but because it is a 0.x version it won't get updated unless it is allowed in the dependency file.

See https://github.com/open-telemetry/opentelemetry-js/releases/tag/v0.12.0 for the breaking changes 

I asked about it before, but it would be great to pin the version number of the provided packages to the opentelemetry version, to ease the matching of the packages.

aabmass:

Part of #188 

Upgrade OpenTelemetry peer dependencies to v0.12.0.

- `Distribution` was removed altogether
- Some other fixes for slight API changes
- I think the codecov drop is flaky since I deleted code?

### DIFF
--- a/packages/opentelemetry-cloud-monitoring-exporter/src/monitoring.ts
+++ b/packages/opentelemetry-cloud-monitoring-exporter/src/monitoring.ts
@@ -140,7 +140,7 @@ export class MetricExporter implements IMetricExporter {
     cb(ExportResult.SUCCESS);
   }
 
-  shutdown(): void {}
+  async shutdown(): Promise<void> {}
 
   /**
    * Returns true if the given metricDescriptor is successfully registered to
--- a/packages/opentelemetry-cloud-trace-exporter/src/trace.ts
+++ b/packages/opentelemetry-cloud-trace-exporter/src/trace.ts
@@ -80,7 +80,7 @@ export class TraceExporter implements SpanExporter {
     resultCallback(result);
   }
 
-  shutdown(): void {}
+  async shutdown(): Promise<void> {}
 
   /**
    * Sends new spans to new or existing traces in the Google Cloud Trace format to the
--- a/packages/opentelemetry-cloud-trace-exporter/test/transform.test.ts
+++ b/packages/opentelemetry-cloud-trace-exporter/test/transform.test.ts
@@ -130,6 +130,7 @@ describe('transform', () => {
   });
 
   it('should drop unknown attribute types', () => {
+    // @ts-expect-error
     readableSpan.attributes.testUnknownType = { message: 'dropped' };
     const result = transformer(readableSpan);
     assert.deepStrictEqual(result.attributes!.droppedAttributesCount, 1);
@@ -172,6 +173,7 @@ describe('transform', () => {
       },
       attributes: {
         testAttr: 'value',
+        // @ts-expect-error
         droppedAttr: {},
       },
     });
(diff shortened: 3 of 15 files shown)


packages/opentelemetry-cloud-trace-propagator/test/CloudPropagator.test.ts
```diff
@@ -75,9 +73,9 @@ describe('CloudPropagator', () => {
     it('should extract context of a sampled span from carrier', () => {
       carrier[X_CLOUD_TRACE_HEADER] =
         'd4cda95b652f4a1592b449d5929fda1b/7929822056569588882;o=1';
-      const extractedSpanContext = getExtractedSpanContext(
```
aabmass, line 78: This function was deleted altogether in favor of using the active span

packages/opentelemetry-cloud-monitoring-exporter/src/monitoring.ts
```diff
@@ -140,7 +140,7 @@ export class MetricExporter implements IMetricExporter {
     cb(ExportResult.SUCCESS);
   }
 
-  shutdown(): void {}
+  async shutdown(): Promise<void> {}
```
dashpole, line 143: just curious, where did this change come from?

dashpole (approved):

packages/opentelemetry-cloud-trace-exporter/test/transform.test.ts
```diff
@@ -130,6 +130,7 @@ describe('transform', () => {
   });
 
   it('should drop unknown attribute types', () => {
+    // @ts-expect-error
```
mayurkale22, line 133: Why do we need this?

packages/opentelemetry-cloud-trace-exporter/src/trace.ts
```diff
@@ -80,7 +80,7 @@ export class TraceExporter implements SpanExporter {
     resultCallback(result);
   }
 
-  shutdown(): void {}
+  async shutdown(): Promise<void> {}
```
mayurkale22, line 83: I think we should add some logic here to set the shutdown flag to true when this method is called and in export return immediately when the flag is set to true. Currently, shutdown operation looks more like a no-op. WDYT? Perhaps, we can do this in separate PR.

packages/opentelemetry-cloud-monitoring-exporter/src/monitoring.ts
aabmass: They changed the interface in the SDK: https://github.com/open-telemetry/opentelemetry-js/pull/1439

packages/opentelemetry-cloud-trace-exporter/src/trace.ts
aabmass: Makes sense, I haven't really look into it yet -- tracking in https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/issues/190

packages/opentelemetry-cloud-trace-exporter/test/transform.test.ts
aabmass: The test is asserting that unsupported attribute types (like object here) are dropped, but it creates a TS error since it is not allowed. In this case, we expect the TS error but want to validate the behavior too

Issue #186, opened 2020-10-15
Update google-auth-library to ^6.1.1


abihani-slb:
There's a security vulnerability (CVSS 3 Score: 7.3 | CVSS 2 Score: 7.5) in the [node-forge](https://www.npmjs.com/package/node-forge) npm package, which is a transitive dependency from google-auth-library:

├─┬ @google-cloud/opentelemetry-cloud-trace-exporter
│ └─┬ google-auth-library
│   └─┬ gtoken
│     └─┬ google-p12-pem
│       └── node-forge

node-forge@0.10.0 has a patch for this vulnerability and [google-auth-library@6.1.1](https://github.com/googleapis/google-auth-library-nodejs/blob/master/CHANGELOG.md) has the patch in it.

abihani-slb:
I see there's already a PR open closing this redundant issue.

aabmass:
Thanks for the report!

aabmass:
> node-forge@0.10.0 has a patch for this vulnerability and [google-auth-library@6.1.1](https://github.com/googleapis/google-auth-library-nodejs/blob/master/CHANGELOG.md) has the patch in it.

Unfortunately, that version of `google-auth-library` doesn't support node8 so we would need to drop node8 or get a backport of the fix.

PR #196, opened 2020-10-30
chore: drop node8, update README.md

aabmass:
Dropping support for node8. Node8 is EOL and hasn't seen a release since last December (almost a year). It is blocking upgrading many packages because their new releases don't support node8. This includes google libraries, e.g. `google-auth-library` which has a security vulnerability (#186) but the fix has only been released for node10+.

Updated the CI to run for versions 10, 12, 14, and 15, which are the [officially supported Node.js versions](https://nodejs.org/en/about/releases/).

The OTel JS SIG is keeping support for node8 in the core packages, because they have very few dependencies. For exporters and instrumentations, which generally have more deps, they will not be uniformly supporting node8.


README.md
```diff
@@ -1,7 +1,11 @@
 # OpenTelemetry Operations Exporters for JavaScript
 
+The packages in this repository support Node.js version 10 and above.
```
nilebox, line 3: Why not 9?
Based on the Circle CI config above, we never checked support for Node 9 and 11 (not sure what are the reasons for that).
Do we need to mention not supporting Node 11 then? Or do we need to add a missing coverage to Circle CI config?

aabmass: Even numbered releases are LTS, so node9 and node11 have been deprecated for a while ([npm releases info](https://nodejs.org/en/about/releases/)):
> Major Node.js versions enter Current release status for six months, which gives library authors time to add support for them. After six months, odd-numbered releases (9, 11, etc.) become unsupported, and even-numbered releases (10, 12, etc.) move to Active LTS status and are ready for general use. LTS release status is "long-term support", which typically guarantees that critical bugs will be fixed for a total of 30 months. Production applications should only use Active LTS or Maintenance LTS releases.

image

aabmass: Updated :)

nilebox (approved):

README.md
nilebox: Thanks!

PR #216, opened 2020-12-17
Mapping http attributes

### LINKED ISSUE #214: http attributes are not mapped accordingly
The trace exporter does not export HTTP attributes to a right format compatible with google (https://cloud.google.com/trace/docs/reference/v2/rest/v2/Attributes)

0.13.0

v14.15.1

### What did you do?
using this code in tracing.js:
```
'use strict';
const {LogLevel, ConsoleLogger} = require('@opentelemetry/core');
const {NodeTracerProvider} = require('@opentelemetry/node');
const {SimpleSpanProcessor} = require('@opentelemetry/tracing');
const {TraceExporter} = require('@google-cloud/opentelemetry-cloud-trace-exporter');
const traceLogger = new ConsoleLogger(LogLevel.DEBUG);
const exporter = new TraceExporter({
  logger: traceLogger
});
const provider = new NodeTracerProvider({logLevel: LogLevel.DEBUG});
provider.addSpanProcessor(new SimpleSpanProcessor(exporter));
provider.register();
```
my package.json:
```
    "@google-cloud/opentelemetry-cloud-trace-exporter": "0.7.0",
    "@opentelemetry/tracing": "0.13.0",
    "@opentelemetry/core": "0.13.0",
    "@opentelemetry/node": "0.13.0",
    "@opentelemetry/plugin-http": "0.13.0",
    "@opentelemetry/plugin-https": "0.13.0",
    "@opentelemetry/plugin-mongodb": "0.11.0"
```
I started my HTTP server and created some GET and POST requests to the server.

I expected to see `/http/method` and `/http/status_text` as labels

### What did you see instead?
But I see 
Screenshot from 2020-12-16 11-08-03

### Related ticket
It seems that the issue is fixed in https://github.com/census-instrumentation/opencensus-node/issues/651 but not in this repository.
### Suggestion:
expose a method on the exporter so I can define my own mapping and change the attribute name of each propery that is important for me. for example like this:
```
const exporter = new TraceExporter({
  customMapping: (fieldName) => {
    const regex = /\./gi;
    return '/'+fieldName.replaceAll(regex, '/');
  }
});
```

sinoohe:
Fixes #214
list of all http attributes are here: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/http.md#common-attributes
I tried to add all of them
also note that in python they did it in this way:
https://github.com/GoogleCloudPlatform/opentelemetry-operations-python/blob/c3772de4cf50950f1d51685ff3bbf0e46ee026df/opentelemetry-exporter-google-cloud/src/opentelemetry/exporter/cloud_trace/__init__.py#L367

### DIFF
--- a/packages/opentelemetry-cloud-trace-exporter/src/transform.ts
+++ b/packages/opentelemetry-cloud-trace-exporter/src/transform.ts
@@ -148,7 +148,7 @@ function valueToAttributeValue(
   }
 }
 
-const HTTP_ATTRIBUTE_MAPPING: { [key: string]: string} = {
+const HTTP_ATTRIBUTE_MAPPING: {[key: string]: string} = {
   'http.method': '/http/method',
   'http.url': '/http/url',
   'http.target': '/http/target',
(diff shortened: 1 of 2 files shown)


sinoohe:
@googlebot I signed it!

PR #232, opened 2021-02-09
chore: upgrade to OTel v0.16.0 packages

aabmass:
- Metrics api moved to separate package `@opentelemetry/api-metrics`
- `setExtractedSpanContext` -> `setSpanContext`
- `getParentSpanContext` -> `getSpanContext`
- Fixed broken status mapping and added tests

### DIFF
--- a/packages/opentelemetry-cloud-trace-exporter/src/transform.ts
+++ b/packages/opentelemetry-cloud-trace-exporter/src/transform.ts
@@ -20,11 +20,13 @@ import {
   AttributeMap,
   Attributes,
   AttributeValue,
+  Code,
   Link,
   LinkType,
   Span,
   Timestamp,
   TruncatableString,
+  Status,
 } from './types';
 import {VERSION} from './version';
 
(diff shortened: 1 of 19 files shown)


packages/opentelemetry-cloud-trace-exporter/src/transform.ts
```diff
@@ -75,6 +77,28 @@ export function getReadableSpanTransformer(
...
+      // TODO: log failed mapping
+      return {code: Code.UNKNOWN, message: status.message};
+    }
+  }
+}
+
+/**
+ * Assert switch case is exhaustive
+ */
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
punya, line 99: I'm surprised that this suppression is necessary - I guess it's because TS thinks that line 89 is unreachable?

punya (approved):
LGTM

packages/opentelemetry-cloud-trace-exporter/src/transform.ts
aabmass: I think the lint is really just local to this function, the `switchValue` param isn't used in the body so it complains. I could also just return the value to avoid disabling lint

PR #242, opened 2021-03-03
chore: update releasing instructions, add check for lockfile v2

aabmass:
- Update releasing.md instructions saying to bump `version.ts` files. This is unfortunate we need to do this and I think I can change it in the future to instead just read from package.json. I believe it was originally done this way for web compat, where you don't necessarily have the package.json file. opentelemetry-js upstream generates these files with a script, which is another option
- Regenerate package locks with `lockfileVersion` 2 again :( Added a CI check that all package locks are using the new version to prevent this in the future

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -20,6 +20,7 @@ jobs:
       - run: npm test
       - run: npm run codecov
         if: matrix.node == 14
+
   lint:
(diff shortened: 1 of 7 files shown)


jsuereth (approved):

aabmass:
Not sure what's with codecov patch, I didn't touch those files so merging anyway

PR #258, opened 2021-04-14
fix: uncaught rejected promise error in MetricExporter's export method

### LINKED ISSUE #254: MetricExporter.export signature doesn't match the base class

@opentelemetry/metrics: ^0.18.2
@google-cloud/opentelemetry-cloud-monitoring-exporter: ^0.9.0

v12.20.2

### What did you do?

The MetricExporter implementation's export definition does not match the base class it's implementing. The base class signature does not return a Promise<void>. However, there is an [open issue](https://github.com/open-telemetry/opentelemetry-js/issues/1023) to support returning a promise. Having these functions not match up causes a Typescript compilation issue and ignoring it causes a dangling promise error. 

I'm referencing line 94 in [packages/opentelemetry-cloud-monitoring-exporter/src/monitoring.ts](https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/blob/103e0a4426560bd9a8cc932af2f03a889f2bd1fe/packages/opentelemetry-cloud-monitoring-exporter/src/monitoring.ts#L94):

```
async export(
    metrics: MetricRecord[],
    cb: (result: ExportResult) => void
  ): Promise<void> {
```

The base class is [MetricExporter from opentelemetry/opentelemetry-metrics](https://github.com/open-telemetry/opentelemetry-js/blob/897c35f497363ee3f3ea4c1e5ad1628e09ec9041/packages/opentelemetry-metrics/src/export/types.ts#L95):

```
export interface MetricExporter {
  /** Exports the list of a given {@link MetricRecord} */
  export(
    metrics: MetricRecord[],
    resultCallback: (result: ExportResult) => void
  ): void;

  /** Stops the exporter. */
  shutdown(): Promise<void>;
}
```

seanachai:
- [X] Make sure to open an issue as a [bug/issue/feature](https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [X] Ensure the tests and linter pass
- [X] Code coverage does not decrease (if any source code was changed)
- [X] Appropriate docs were updated (if necessary)

Fixes #254


seanachai:
@googlebot I signed it!

packages/opentelemetry-cloud-monitoring-exporter/src/monitoring.ts
```diff
@@ -93,7 +93,13 @@ export class MetricExporter implements IMetricExporter {
     cb: (result: ExportResult) => void
   ): Promise<void> {
     if (this._projectId instanceof Promise) {
-      this._projectId = await this._projectId;
+      try {
```
aabmass, line 96: Nice! I think the await on L114 could also throw. Maybe we could do something like this to generalize:

```ts
  async _exportAsync(
    metrics: MetricRecord[],
  ): Promise<ExportResult> {
    // this comes from renaming and updating params for the existing export() function
    // and instead return the ExportResult
  }
  
  export(
    metrics: MetricRecord[],
    cb: (result: ExportResult) => void
  ): void {
    _exportAsync(metrics)
      .then(cb)
      .except(err => {
        diag.error(err.message);
        cb(/* map err to ExportResult */);
      });
  }
```

seanachai: Oh wow yeah great catch. I totally missed the await on L114. Great idea on wrapping the async calls in their own function. I'll get on that asap.

seanachai: @aabmass apologies for the delay here. Lmk how this looks. I added some comments for clarity. I'm happy to add some more tests too if necessary.

aabmass: No worries. I think you can remove this new try block since it will be caught in the `.catch()` above.

packages/opentelemetry-cloud-monitoring-exporter/src/monitoring.ts
```diff
@@ -92,14 +93,38 @@ export class MetricExporter implements IMetricExporter {
     metrics: MetricRecord[],
     cb: (result: ExportResult) => void
   ): Promise<void> {
+    return this._exportAsync(metrics)
```
aabmass, line 96: This function doesn't need to be async anymore

packages/opentelemetry-cloud-monitoring-exporter/src/monitoring.ts
```diff
@@ -92,14 +93,38 @@ export class MetricExporter implements IMetricExporter {
...
+        return cb({code: ExportResultCode.FAILED, error: err});
+      });
+  }
+
+  async shutdown(): Promise<void> {}
+
+  /**
+   * Asnyc wrapper for the {@link export} implementation.
+   * Saves the current values of all exported {@link MetricRecord}s so that
+   * they can be pulled by the Google Cloud Monitoring backend.
```
aabmass, line 109: Just noticed this original description doesn't really make sense, there is no pulling. Can we just make it "Writes MetricRecords to Cloud Monitoring" or similar?

packages/opentelemetry-cloud-monitoring-exporter/test/monitoring.test.ts
```diff
@@ -135,6 +135,53 @@ describe('MetricExporter', () => {
...
+      let unhandledPromiseRejectionEvent = false;
+      process.on('unhandledRejection', error => {
+        unhandledPromiseRejectionEvent = true;
+      });
+
+      const result = await new Promise<ExportResult>(resolve => {
+        exporter.export(records, result => {
+          resolve(result);
+        });
+      });
```
aabmass, line 180: Since we are still returning the promise, this would be equivalent?

```suggestion
      const result = await exporter.export(records, () => {});
```

packages/opentelemetry-cloud-monitoring-exporter/test/monitoring.test.ts
```diff
@@ -135,6 +135,53 @@ describe('MetricExporter', () => {
...
+
+      assert.strictEqual(unhandledPromiseRejectionEvent, false);
```
aabmass, line 182: Did you find this test failed before the fix? I'm not sure if `unhandledRejection` handler is guaranteed to be called before this assertion is made

aabmass (commented):
LGTM, just a few small comments

packages/opentelemetry-cloud-monitoring-exporter/test/monitoring.test.ts
seanachai: I wanted to make sure there was a regression test covering what we're improving. I ran this without my changes and it triggered a failure. Admittedly, I haven't used this in practice so I can't say if this is a good practice or not.

packages/opentelemetry-cloud-monitoring-exporter/test/monitoring.test.ts
seanachai: Oh yeah this makes sense. I was just following previous test patterns.

packages/opentelemetry-cloud-monitoring-exporter/src/monitoring.ts
seanachai: Which function?

aabmass: Sorry, the `export()` function, you can just remove the `async` since it doesn't await anymore

packages/opentelemetry-cloud-monitoring-exporter/test/monitoring.test.ts
aabmass: gotcha. either way is fine with me, feel free to resolve

seanachai: Sounds good 👍🏻 Apologies for the major delay here. May has been quite a month for me. I'll get to wrapping this up by EoW.

packages/opentelemetry-cloud-monitoring-exporter/test/monitoring.test.ts
```diff
@@ -243,7 +289,11 @@ describe('MetricExporter', () => {
       await meter.collect();
       const records2 = meter.getProcessor().checkPointSet();
 
-      await exporter.export(records2, () => {});
+      const result2 = await new Promise<ExportResult>(resolve => {
```
seanachai, line 292: I believe removing the async from the export function made these changes necessary.

seanachai:
@aabmass made those latest changes. Lmk what you think 🍻

aabmass (approved):
Nice, thank you for the PR!

aabmass:
@seanachai could you fix the lint warnings? I think the codecov patch check will be fixed if you rebase on master.

seanachai:
> @seanachai could you fix the lint warnings? I think the codecov patch check will be fixed if you rebase on master.

Yup, will do 👍

seanachai:
Thanks @aabmass!

PR #268, opened 2021-05-22
feat: stop adding project id to span attributes

### LINKED ISSUE #267: Stop adding project id to all spans
This happens here:

https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/blob/0fb13d7be54f8175384b26bbb02bdedff5e39824/packages/opentelemetry-cloud-trace-exporter/src/transform.ts#L43

This is unnecessary and doesn't follow our "spec" of GCT exporters.

aabmass:
Fixes #267 

BREAKING CHANGE: spans will no longer have project id attribute

### DIFF
--- a/packages/opentelemetry-cloud-monitoring-exporter/package-lock.json
+++ b/packages/opentelemetry-cloud-monitoring-exporter/package-lock.json
@@ -1,10834 +1,10834 @@
 {
-  "name": "@google-cloud/opentelemetry-cloud-monitoring-exporter",
-  "version": "0.9.0",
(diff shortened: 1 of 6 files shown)


dashpole (approved):

jsuereth (approved):
Is this a breaking change, and do we need a changelog for this project or do you just curate that when you cut a release?

aabmass:
> Is this a breaking change, and do we need a changelog for this project or do you just curate that when you cut a release?

Yes it's breaking and the latter

PR #281, opened 2021-06-16
feat: add basicTrace test scenario

aabmass:


aabmass:
Squashed into another PR

PR #282, opened 2021-06-16
feat: separate resource mapping utils into separate package

aabmass:
Refactor the code to map OTel resource into a monitored resource into a new separate package `@google-cloud/opentelemetry-resource-util`. Then, I can use this package in the trace exporter to create `g.co/r/` span attributes from the OTel monitored resource.


packages/opentelemetry-resource-util/src/index.ts
```diff
@@ -0,0 +1,80 @@
...
+ * return the global resource. The only currently supported monitored resources
+ * are gce_instance and aws_ec2_instance.
+ * @param resource
+ * @param projectId
+ */
+export function mapOtelResourceToMonitoredResource(
+  resource: Resource,
+  projectId: string
+): MonitoredResource {
+  const cloudProvider = `${
```
punya, line 37: Instead of string-interpolating this, could we say

```ts
const cloudProvider = resource.attributes[ResourceAttributes.CLOUD_PROVIDER] as string | undefined;
```

Otherwise the undefined case gets mapped to the string `"undefined"` which is more likely to trip us up later.

packages/opentelemetry-resource-util/src/index.ts
```diff
@@ -0,0 +1,80 @@
...
+          ResourceAttributes.CLOUD_ACCOUNT_ID
+        ] as string,
+      },
+    };
+  }
+
+  // fallback or if the resource is missing label values
+  if (
+    monitoredResource === undefined ||
+    hasUndefinedValue(monitoredResource.labels)
```
punya, line 70: Optional and stylistic, but I would find this easier to understand if we wrote out the undefined checks explicitly in the branches above. Something like

```ts
switch (cloudProvider) {
  case 'gcp':
    if (ResourceAttributes.HOST_ID not in resource.attributes || ResourceAttributes.CLOUD_AVAILABILITY_ZONE not in resource.attributes) return;
    return { type: GCP_GCE_INSTANCE, labels: { ... } };
  case 'aws':
    // ...
}
```

In particular, when I look at the `aws` branch, I see that the `region` label would never be undefined because of string interpolation -- is that what we actually want?

packages/opentelemetry-resource-util/src/index.ts
aabmass: good catch

packages/opentelemetry-resource-util/src/index.ts
aabmass: no, the string interpolation is a bug. let me rewrite this a little more clearly 👍

punya (approved):

PR #293, opened 2021-06-23
feat(trace-exporter)!: send span kind to cloud trace

aabmass:
BREAKING CHANGE: add span kind is now sent with traces.

### DIFF
--- a/packages/opentelemetry-cloud-trace-exporter/src/types.ts
+++ b/packages/opentelemetry-cloud-trace-exporter/src/types.ts
@@ -14,6 +14,8 @@
 
 import {Metadata} from '@grpc/grpc-js';
 
+// extends cloudtrace_v2.Schema$Span
+
 export interface Span {
   name?: string;
   spanId?: string;
@@ -30,6 +32,7 @@ export interface Span {
   status?: Status;
   sameProcessAsParentSpan?: BoolValue;
   childSpanCount?: number;
+  spanKind?: SpanKind;
 }
 
 export interface Timestamp {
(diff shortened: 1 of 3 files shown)


packages/opentelemetry-cloud-trace-exporter/src/types.ts
```diff
@@ -14,6 +14,8 @@
 
 import {Metadata} from '@grpc/grpc-js';
 
+// extends cloudtrace_v2.Schema$Span
```
punya, line 17: Why is this a comment instead of actually extending that interface?

packages/opentelemetry-cloud-trace-exporter/src/types.ts
```diff
@@ -166,3 +169,35 @@ export enum Code {
   OK = 0,
   UNKNOWN = 2,
 }
+
+/**
+ * See https://github.com/googleapis/googleapis/blob/master/google/devtools/cloudtrace/v2/trace.proto#L182
```
punya, line 174: Could you reference a tagged version instead of master?

punya (approved):
Small nits, but looks good overall.

packages/opentelemetry-cloud-trace-exporter/src/types.ts
aabmass: Oops I just forgot to delete that. I was trying exactly that but the generated types are not very useful (`string | null`) so I'd rather copy in an enum. It is `string | null` because JS `number` can't represent large integers, but that is no concern here. They get serialized just fine from the enum

packages/opentelemetry-cloud-trace-exporter/src/types.ts
aabmass: There are no useful tags in there unfortunately. I can do a permarlink from the commit (its super long but oh well)

weyert:
This PR sounds interesting @aabmass Does this in any change the way the span is represented in Google Trace UI?

aabmass:
I don't think the UI shows span kind at the moment, but it is definitely saved in the backend

PR #294, opened 2021-06-24
feat(e2e-testing): implement complexTrace e2e scenario

aabmass:


aabmass:
Not sure why codecov is complaining since it should be ignoring this path. I guess it still contributes toward the overall project? https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/blob/1ca9844158eee663c513c6f4da748101a4c52b69/codecov.yaml#L15-L16

tbarker25 (approved):

PR #318, opened 2021-08-24
fix: always use forward slash when creating MetricDescriptor name and type

### LINKED ISSUE #317: Cloud Monitoring Exporter creates invalid MetricDescriptor types on Windows
0.24.0

16.4

### What did you do?

Use the usage example at https://www.npmjs.com/package/@google-cloud/opentelemetry-cloud-monitoring-exporter on Windows.

Metrics would be exported and viewable on the Google Cloud Monitoring UI.

### What did you see instead?

No metrics are exported.

When you attach a DiagConsoleLogger to OpenTelemetry, you can see that the request to create a MetricDescriptor fails with the message: `Field metricDescriptor.type had an invalid value of "custom.googleapis.com\\opentelemetry\\metric_name": The metric type must be a URL-formatted string with a domain and non-empty path.`.

Since `path.join()` is used to create the type, all the forward slashes are replaced with back slashes on Windows, rendering the type invalid.

### Additional context
Works as expected on Linux.

BlazingAsher:
- [ ] Make sure to open an issue as a [bug/issue/feature](https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [ ] Ensure the tests and linter pass

Fixes #317

### DIFF
--- a/packages/opentelemetry-cloud-monitoring-exporter/src/transform.ts
+++ b/packages/opentelemetry-cloud-monitoring-exporter/src/transform.ts
@@ -60,12 +60,12 @@ export function transformMetricDescriptor(
 
 /** Transforms Metric type. */
 function transformMetricType(metricPrefix: string, name: string): string {
-  return path.join(metricPrefix, name);
+  return path.posix.join(metricPrefix, name);
 }
 
 /** Transforms Metric display name. */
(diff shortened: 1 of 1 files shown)


aabmass:
Thank you for the fix!

PR #330, opened 2021-09-22
chore: fix flaky/slow test in Cloud Trace exporter

### LINKED ISSUE #160: Mocking of GCP environment is insufficient
The mocking here is not sufficient to prevent the `google-auth-library` from detecting GCE in some cases.
https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/blob/aa8a6441c36931ca40b438a61c8060da2ad7a426/packages/opentelemetry-cloud-trace-exporter/test/exporter.test.ts#L62-L68

On my machine, the test code still detected the GCP project I am logged into with gcloud, and so it never calls the metadata server. I think its because I ran `gcloud auth application-default login`, which creates `~/.config/gcloud/application_default_credentials.json`.

aabmass:
Fixes #160 

This test was previously using `nock` to fake the GCP metadata server and fool google-auth-library when detecting the project ID. This didn't always work (e.g. google-auth-library might find project ID in a gcloud SDK installation) and was testing implementation details of google-auth-library.

Instead, we can just mock the whole `GoogleAuth` class away since we are only testing the exporter functionality.


packages/opentelemetry-cloud-trace-exporter/test/exporter.test.ts
```diff
@@ -47,27 +42,30 @@ describe('Google Cloud Trace Exporter', () => {
...
-      assert(exporter);
-      return (exporter['_projectId'] as Promise<string>).then(id => {
-        assert.deepStrictEqual(id, 'not-real');
-        gcpMock.done();
-      });
-    }).timeout(10000);
+      assert.ok(exporter);
+      const id = (await exporter['_projectId']) as string;
+      assert.ok(getProjectIdFake.calledOnce);
+      assert.deepStrictEqual(id, 'fake-project-id');
```
aabmass, line 67:
```suggestion
      assert.strictEqual(id, 'fake-project-id');
```

PR #359, opened 2021-11-24
Add compatibility for new metricKinds

### LINKED ISSUE #358: Incompatible with opentelemetry-js 0.27

0.27

16

### What did you do?

I am creating an observableGauge (former valueObserver). When sending requests to gcp stackdriver, I get a response that the metricKind is unknown.

I expected to see that the request is working.

### What did you see instead?

The request failing.

### Additional context

Hello there,

we have recently updated our version of opentelemetry-js to 0.27.0. While doing that, a lot of stuff got renamed, for example the valueObserver which we were using. It is now called observableGauge and has a different METRIC_KIND set. This kind is not supported yet by this library as it seems.

ferencbeutel4711:
Fixes #358

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,7 +5,6 @@
   "requires": true,
   "packages": {
     "": {
-      "name": "opentelemetry-operations-js",
       "version": "0.4.0",
       "hasInstallScript": true,
       "license": "Apache-2.0",
--- a/packages/opentelemetry-cloud-monitoring-exporter/package-lock.json
+++ b/packages/opentelemetry-cloud-monitoring-exporter/package-lock.json
@@ -15,10 +15,10 @@
 			},
 			"devDependencies": {
 				"@opentelemetry/api": "1.0.3",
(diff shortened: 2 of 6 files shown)


aabmass (commented):
I don't think these changes have been released yet, right? It's probably best to wait until the JS SDK is in a working state, since it is being re-implemented for the updated specification.

aabmass:
OK I took another look and I see the renames were done. I think you need to upgrade the `@opentelemetry/api-metrics` version here https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/blob/28ee4ada202122b09ffcd4566cc5ff184e0ecf08/packages/opentelemetry-cloud-monitoring-exporter/package.json#L43

aabmass:
@ferencbeutel4711 can you sign the CLA and fix the conventional commits issue? I updated the PR so otherwise it should be good to go.

aabmass:
Superseded by #365

PR #393, opened 2022-04-14
feat: Splitting metrics CUSTOM_OPENTELEMETRY_DOMAIN from prefix

### LINKED ISSUE #386: Could Metrics Exporter prefix option be split into two separate options?
The MetricExporter accepts a single string for the prefix option. The class is using prefix to define both _metricPrefix and _displayNamePrefix. If prefix is not set the class provides a different value for _metricPrefix than _displayNamePrefix:

https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/blob/main/packages/opentelemetry-cloud-monitoring-exporter/src/monitoring.ts#L52To#L54
```
static readonly DEFAULT_DISPLAY_NAME_PREFIX: string = 'OpenTelemetry';
static readonly CUSTOM_OPENTELEMETRY_DOMAIN: string ='custom.googleapis.com/opentelemetry';
```
Which are used later: 
https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/blob/main/packages/opentelemetry-cloud-monitoring-exporter/src/monitoring.ts#L61To#L65
```
constructor(options: ExporterOptions = {}) {
    this._metricPrefix =
      options.prefix || MetricExporter.CUSTOM_OPENTELEMETRY_DOMAIN;
    this._displayNamePrefix =
      options.prefix || MetricExporter.DEFAULT_DISPLAY_NAME_PREFIX;
```

I would like to be able to set _displayNamePrefix using a different variable. This would help me better filter metrics in the UI.

In my testing I have found that in order for the metrics to be displayed in the Metrics Explorer UI metricPrefix must use the domain: custom.googleapis.com/{something} where something is a custom name. For example, I might want to set {something} to the name of my application.  That way I could filter out that application in the UI. Given that prefix is used as the value for the two variables, to implement the application name I would have to set prefix to: custom.googleapis.com/applicationName. Which means that the display name includes that along with the metric you want to observe.  This looks terrible. 

Unless I am missing something, it would seem like this could be an easy feature to implement. Perhaps prefix could be split into something more descriptive like:

- displayNamePrefix
- customOTDomain

I don't think this would break anything. 

Would there be a reason why metricPrefix would not point to the custom.googleapis.com URL?

dwashko:
Splitting metric option prefix into displayNamePrefix and customOTDomain.

Fixes #386

### DIFF
--- a/packages/opentelemetry-cloud-monitoring-exporter/README.md
+++ b/packages/opentelemetry-cloud-monitoring-exporter/README.md
@@ -31,14 +31,16 @@ const meter = new MeterProvider({
 const counter = meter.createCounter('metric_name');
 counter.add(10, { [key]: 'value' });
 ```
-## Setting Custom Display Name
+## Setting Custom Display Name - Prefix
 
 ```js
 const exporter = new MetricExporter({
-  customDisplayName: 'myApplication',
+  prefix: 'myApplication',
 }
 ```
-When viewing the metric you should be able to filter on the customDisplayName value (e.g.; myApplication). 
+When viewing the metric you should be able to filter on the prefix value (e.g.; myApplication). 
+
+In addition to setting a custom prefix you can specify a custom OpenTelemetry Domain using the customOTDomain option. The default value is custom.googleapis.com/opentelemetry.
 
 ##  Viewing your metrics:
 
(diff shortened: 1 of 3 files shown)


aabmass:
@dwashko thanks for the PR. See my comment on https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/issues/386#issuecomment-1331663104, but I think I have improved the default behavior where this configuration shouldn't be necessary. Please follow re-open the other issue or open a new one if this still doesn't work for you.

PR #403, opened 2022-05-10
Add apiEndpoint option to cloud monitoring exporter

dashpole:
The underlying client library doesn't expose apiEndpoint like the trace one does, so I've named it differently (rootUrl instead of apiEndpoint) accordingly.

### DIFF
--- a/packages/opentelemetry-cloud-monitoring-exporter/src/monitoring.ts
+++ b/packages/opentelemetry-cloud-monitoring-exporter/src/monitoring.ts
@@ -20,7 +20,7 @@ import {
 import {ExportResult, ExportResultCode, VERSION} from '@opentelemetry/core';
 import {ExporterOptions} from './external-types';
 import {GoogleAuth, JWT} from 'google-auth-library';
-import {google} from 'googleapis';
+import {google, monitoring_v3} from 'googleapis';
 import {transformMetricDescriptor, createTimeSeries} from './transform';
 import {TimeSeries} from './types';
 import {partitionList} from './utils';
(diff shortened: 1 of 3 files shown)


packages/opentelemetry-cloud-monitoring-exporter/src/monitoring.ts
```diff
@@ -72,6 +72,11 @@ export class MetricExporter implements IMetricExporter {
       scopes: ['https://www.googleapis.com/auth/cloud-platform'],
     });
 
+    this._monitoring = google.monitoring({
+      version: 'v3',
+      rootUrl: options.rootUrl || 'https://monitoring.googleapis.com/',
```
aabmass, line 77: Should we make this the same as trace exporter with `apiEndpoint` and just add the `https://` prefix on our own?

dashpole: It still wouldn't include the port (e.g. the apiEndpoint on trace includes :443 as well).

dashpole: I changed it to add the https:// prefix, and to include the port

dashpole: And I changed it to be called "apiEndpoint" to match the trace client

aabmass (approved):

packages/opentelemetry-cloud-monitoring-exporter/src/monitoring.ts
aabmass: LGTM

PR #411, opened 2022-09-29
chore: increase gce integration test timeout

aabmass:
It is timing out waiting for the container to start https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/pull/371/checks?check_run_id=8621886819

If this gets much longer, we may need to investigate. I had to increase it for operations-python repo as well, but operations-go doesn't seem to have a similar problem. It could be the docker image is too large.


dashpole:
operations-go has had the same problem.  I've hit it a few times

dashpole (approved):

PR #409, opened 2022-09-21
feat(cloud-monitoring-exporter): Support Distribution exporting

markrzen:
Thank you for opening a Pull Request! Before submitting your PR, there are a few things you can do to make sure it goes smoothly:
- [x] Make sure to open an issue as a [bug/issue/feature](https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [x] Ensure the tests and linter pass
- [x] Code coverage does not decrease (if any source code was changed)
- [x] Appropriate docs were updated (if necessary)

Fixes #101

PR #428, opened 2022-09-30
feat(cloud-monitoring-exporter): Support Distribution exporting

### LINKED ISSUE #101: Add distribution and histogram metric support

markrzen:

Fixes #101 
Replaces #409


packages/opentelemetry-cloud-monitoring-exporter/src/transform.ts
```diff
@@ -73,20 +76,27 @@ function transformMetricKind(kind: OTMetricKind): MetricKind {
...
-function transformValueType(valueType: OTValueType): ValueType {
+function transformValueType(
+  valueType: OTValueType,
+  metricKind: OTMetricKind
+): ValueType {
+  // Stackdriver ValueType is not a 1 to 1 with OTel ValueType
   if (valueType === OTValueType.DOUBLE) {
+    if (metricKind === OTMetricKind.HISTOGRAM) {
+      return ValueType.DISTRIBUTION;
+    }
```
markrzen, line 99: Histograms are both a OTValueType.DOUBLE and a OTMetricKind.HISTOGRAM.

packages/opentelemetry-cloud-monitoring-exporter/src/transform.ts
```diff
@@ -144,6 +157,7 @@ function transformPoint(
   switch (metricDescriptor.metricKind) {
     case OTMetricKind.COUNTER:
     case OTMetricKind.OBSERVABLE_COUNTER:
+    case OTMetricKind.HISTOGRAM:
```
markrzen, line 160: Histograms require a startDate.

markrzen:
@dashpole OK, CLA signed. Comments copied over from the original PR #409, which got closed out when I sync'd back to my fork.

markrzen:
@dashpole Are y'all trying to maintain support for no longer supported versions of Node? 10.x and 12.x?
If so, I will need to revert the mocha upgrades.

markrzen:
@dashpole Went ahead and dropped mocha/ts-mocha back down to 9.x to allow for Node 10.x support.

dashpole:
I'm not sure.  @aabmass should know.  I'm just here to poke the test runner. :)

markrzen:
Tests passed. 👍 👍

aabmass:
@markrzen thanks for the PR. It looks like this is still building against [@opentelemetry/sdk-metrics-base](https://www.npmjs.com/package/@opentelemetry/sdk-metrics-base) which is deprecated. I think the first step would be upgrading to a newer metrics release on the new packages since so much has changed upstream, and adapting the existing code. I think this would be a sizeable amount of work which i definitely don't eoxec

I apologize for not getting around to this work, I haven't had much time to maintain this work.

packages/opentelemetry-cloud-monitoring-exporter/src/transform.ts
```diff
@@ -73,20 +76,27 @@ function transformMetricKind(kind: OTMetricKind): MetricKind {
   switch (kind) {
     case OTMetricKind.COUNTER:
     case OTMetricKind.OBSERVABLE_COUNTER:
+    case OTMetricKind.HISTOGRAM:
```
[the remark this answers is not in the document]
aabmass: That's right. Btw, we've implemented some canonical mappings from OTel to GCM in the collector exporter. We hope to implement the same logic in the language exporters.

markrzen:
Moved to using Prometheus for easier integration.

PR #442, opened 2022-10-27
feat: export array type attributes

### LINKED ISSUE #440: Add support for array attributes
*Is your feature request related to a problem? Please describe.*
OpenTelemetry Spans support attributes with array values, these attributes are currently not exported to CloudTrace. This is not notified of in any way (at least not that I can identify), which leads to application developers suspecting faults in their use of the library.

*Describe the solution you'd like*
[Code that transforms Spans](https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/blob/main/packages/opentelemetry-cloud-trace-exporter/src/transform.ts#L204-L218) should support array types from the [list of allowed types](https://github.com/open-telemetry/opentelemetry-js-api/blob/main/src/common/Attributes.ts#L31-L37). This might require changes in CloudTrace APIs, which I can also file a feature request for, presumably [here](https://issuetracker.google.com/issues?q=status:open%20componentid:187216&s=created_time:desc)?

*Describe alternatives you've considered*
There are workarounds, like concatenating values into a singular string, but that's hacky, especially as OpenTelemetry Spans do allow for array types of attributes.

henrinormak:
- [X] Make sure to open an issue as a [bug/issue/feature](https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [X] Ensure the tests and linter pass
- [X] Code coverage does not decrease (if any source code was changed)
- [ ] Appropriate docs were updated (if necessary) - not sure if there are any docs that need updating

Fixes #440 (sort of, at least until proper API level support for array types is added to CloudTrace)

### DIFF
--- a/packages/opentelemetry-cloud-trace-exporter/src/external-types.ts
+++ b/packages/opentelemetry-cloud-trace-exporter/src/external-types.ts
@@ -47,7 +47,7 @@ export interface TraceExporterOptions {
    */
   apiEndpoint?: string;
   /**
-   * If enabled, array type Span attributes are stringified before
+   * If enabled, array type Span attributes are JSON stringified before
    * exporting. Default: false, which means array attributes are simply
    * ignored and not exported
    */
--- a/packages/opentelemetry-cloud-trace-exporter/src/external-types.ts
+++ b/packages/opentelemetry-cloud-trace-exporter/src/external-types.ts
@@ -46,6 +46,12 @@ export interface TraceExporterOptions {
    * cloudtrace.googleapis.com.
    */
   apiEndpoint?: string;
+  /**
+   * If enabled, array type Span attributes are stringified before
+   * exporting. Default: false, which means array attributes are simply
+   * ignored and not exported
+   */
+  stringifyArrayAttributes?: boolean;
 }
 
 export interface Credentials {
--- a/packages/opentelemetry-cloud-trace-exporter/src/transform.ts
+++ b/packages/opentelemetry-cloud-trace-exporter/src/transform.ts
@@ -37,24 +37,33 @@ const AGENT_LABEL_VALUE = `opentelemetry-js ${CORE_VERSION}; google-cloud-trace-
 
 export function getReadableSpanTransformer(
   projectId: string,
-  resourceFilter?: RegExp | undefined
+  resourceFilter?: RegExp | undefined,
+  stringifyArrayAttributes?: boolean
 ): (span: ReadableSpan) => Span {
   return span => {
     // @todo get dropped attribute count from sdk ReadableSpan
     const attributes = mergeAttributes(
-      transformAttributes({
-        ...span.attributes,
-        [AGENT_LABEL_KEY]: AGENT_LABEL_VALUE,
-      }),
+      transformAttributes(
+        {
+          ...span.attributes,
+          [AGENT_LABEL_KEY]: AGENT_LABEL_VALUE,
+        },
+        stringifyArrayAttributes
+      ),
       // Add in special g.co/r resource labels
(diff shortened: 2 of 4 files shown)


henrinormak:
Amended the change I made to make this feature opt-in via config passed to the constructor of the exporter.

packages/opentelemetry-cloud-trace-exporter/src/transform.ts
```diff
@@ -222,8 +272,15 @@ function stringToTruncatableString(value: string): TruncatableString {
...
 function valueToAttributeValue(
-  value: string | number | boolean
-): AttributeValue {
+  value:
+    | string
+    | number
+    | boolean
+    | Array<string | null | undefined>
+    | Array<number | null | undefined>
+    | Array<boolean | null | undefined>,
```
aabmass, line 281: Can use `AttributeValue` type alias from OTel https://github.com/open-telemetry/opentelemetry-js/blob/e133f3f5c3d9467c5c96df14a5fdd6c43706d450/api/src/common/Attributes.ts#L31

packages/opentelemetry-cloud-trace-exporter/src/transform.ts
```diff
@@ -233,7 +290,12 @@ function valueToAttributeValue(
     case 'string':
       return {stringValue: stringToTruncatableString(value)};
     default:
-      return {};
+      if (stringifyArrayAttributes) {
+        return {stringValue: stringToTruncatableString(JSON.stringify(value))};
```
aabmass, line 294: It would be good to drop `undefined` explicitly, as `JSON.stringify()` ts type is incorrect https://github.com/microsoft/TypeScript/issues/18879

packages/opentelemetry-cloud-trace-exporter/src/external-types.ts
```diff
@@ -46,6 +46,12 @@ export interface TraceExporterOptions {
    * cloudtrace.googleapis.com.
    */
   apiEndpoint?: string;
+  /**
+   * If enabled, array type Span attributes are stringified before
```
aabmass, line 50:
```suggestion
   * If enabled, array type Span attributes are JSON stringified before
```

aabmass (commented):
Thanks for the PR, few questions otherwise LGTM

packages/opentelemetry-cloud-trace-exporter/src/transform.ts
```diff
@@ -198,20 +226,42 @@ function transformResourceToAttributes(
...
   for (const [key, value] of Object.entries(attributes)) {
-    switch (typeof value) {
-      case 'number':
-      case 'boolean':
-      case 'string':
-        out[key] = valueToAttributeValue(value);
-        break;
-      default:
-        break;
+    if (!isAttributeValue(value)) {
```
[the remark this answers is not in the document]
henrinormak: Removed, it was a leftover from initial changes

packages/opentelemetry-cloud-trace-exporter/src/transform.ts
henrinormak: Do you mean dropping undefined from the array values? Because the typing is such that the value itself cannot be undefined, `transformAttributeValues` also specifically checks for that ahead of calling `valueToAttributeValue`.

Values in the array could be though, they can also be nulls, should both of those be dropped from arrays?

aabmass: Ah sorry, I missed that you already filter the undefined out in the function  above 👍

aabmass (approved):
Thanks for contributing!

Issue #379, opened 2022-03-09
Is the Google Cloud Monitoring Exporter Actually Supported


mamidon:
Please answer these questions before submitting a bug report.

### What version of OpenTelemetry are you using?
"@opentelemetry/sdk-metrics-base": "0.27.0",
"@google-cloud/opentelemetry-cloud-monitoring-exporter": "^0.14.0",

### What version of Node are you using?
v12.22.7

### What did you do?
If possible, provide a recipe for reproducing the error.
I followed the example at [the readme](https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/blob/main/packages/opentelemetry-cloud-monitoring-exporter/README.md) which has it's own minor syntax error.

### What did you expect to see?
I expect code equivalent to the example, with a minor syntax fix, to compile successfully.

### What did you see instead?
Compilation errors
```
../../node_modules/@google-cloud/opentelemetry-cloud-monitoring-exporter/build/src/monitoring.d.ts(26,5): error TS2416: Property 'export' in type 'MetricExporter' is not assignable to the same property in base type 'MetricExporter'.
Type '(metrics: import("/.../node_modules/@opentelemetry/sdk-metrics-base/build/src/export/types").MetricRecord[], resultCallback: (result: import("/.../node_modules/@opentelemetry/core/build/src/ExportResult").ExportResult) => void) => void' is not assignable to type '(metrics: import("/.../node_modules/@opentelemetry/sdk-metrics-base/build/src/export/types").MetricRecord[], resultCallback: (result: import("/.../node_modules/@opentelemetry/sdk-metrics-base/node_modules/@opentelemetry/core/build/src/ExportResult").E...'.
     Types of parameters 'resultCallback' and 'resultCallback' are incompatible.
       Types of parameters 'result' and 'result' are incompatible.
         Type 'import("/.../node_modules/@opentelemetry/core/build/src/ExportResult").ExportResult' is not assignable to type 'import("/.../node_modules/@opentelemetry/sdk-metrics-base/node_modules/@opentelemetry/core/build/src/ExportResult").ExportResult'.
```

### Additional context
The code I've written is basically what's in the readme
```
import {MeterProvider} from '@opentelemetry/sdk-metrics-base';
import { MetricExporter } from '@google-cloud/opentelemetry-cloud-monitoring-exporter';

const exporter = new MetricExporter();

// Register the exporter
const meter = new MeterProvider({
    exporter,
[...]

PR #446, opened 2022-11-19
fix(cloud-monitoring-exporter)!: update to stable OTel metrics version

aabmass:
Fixes https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/issues/379
Fixes https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/issues/129
Fixes https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/issues/101

Lots of changes to upgrade to the latest (stable) version of OTel metrics.

I removed all of `packages/opentelemetry-cloud-monitoring-exporter/test/transform.test.ts` in favor of snapshot/golden tests I am adding in the next PR: #447.

Note codecov is failing because of the deleted tests


.github/workflows/ci.yaml
```diff
@@ -9,12 +9,23 @@ jobs:
...
+            npm-version: ^8
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
         with:
           node-version: ${{ matrix.node }}
       - run: node --version
+      # TODO: remove once node 14 is dropped
+      - name: Upgrade to latest npm to support lockfile v2
+        run: npm install -g npm@${{ matrix.npm-version }}
```
aabmass, line 28: Without this, npm was failing for older node versions which come with npm<7 and claiming `@opentelemetry/sdk-node` wasn't installed.

aabmass: Moved these changes to https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/pull/448, will rebase after it's merged.

packages/opentelemetry-cloud-monitoring-exporter/test/monitoring.test.ts
```diff
@@ -177,48 +171,42 @@ describe('MetricExporter', () => {
         unhandledPromiseRejectionEvent = true;
       });
 
-      await exporter.export(records, () => {});
+      await new Promise<ExportResult>(resolve => {
```
punya, line 174: Should we use util.promisify instead?

.github/workflows/ci.yaml
aabmass: Looks like you reviewed the other PR. The comment is correct, it's that node14 comes with npm 6. However npm@latest still supports node14, so the default value in the matrix `latest` works fine

.github/workflows/ci.yaml
```diff
@@ -9,12 +9,23 @@ jobs:
     strategy:
       matrix:
         node: [10, 12, 14, 16]
+        include:
+          # use latest npm by default
+          - npm-version: latest
+          # overrides for node 10 and 12
```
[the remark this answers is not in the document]
aabmass: Still needed for node14 see comment above.

packages/opentelemetry-cloud-monitoring-exporter/test/monitoring.test.ts
aabmass: I actually tried that, but it won't work correctly because the signature used in the upstream SDK doesn't follow the [error first convention](https://nodejs.org/dist/latest-v8.x/docs/api/util.html#util_util_promisify_original:~:text=Takes%20a%20function%20following%20the%20common%20error%2Dfirst%20callback%20style%2C%20i.e.%20taking%20a%20(err%2C%20value)%20%3D%3E%20...%20callback%20as%20the%20last%20argument%2C%20and%20returns%20a%20version%20that%20returns%20promises.):

```ts
function export(
  metrics: ResourceMetrics,
  resultCallback: (result: {code: ExportResultCode, error?: Error}) => void
): void;
```

since there is only one argument `result` to the callback, it is interpreted as the error. IOW, typescript selects [this overload](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/adc1b6dd3233c3478e0744c0112355b8e2b431a1/types/node/util.d.ts#L1009):

```ts
export function promisify(fn: (callback: (err?: any) => void) => void): () => Promise<void>;
```

punya (approved):

aabmass:
Going to merge with codecov checks failing since tests are added in the subsequent PRs

PR #448, opened 2022-11-19
chore: set npm major version per node version in CI

### LINKED ISSUE #364: Set npm major version in CI
See how it was done in OpenCensus: https://github.com/census-instrumentation/opencensus-node/blob/1c12e6fda5a9f38e36089066f431f9c47518e94c/.github/workflows/ci.yaml#L13-L22

This is needed because older versions of node come with npm<7. Npm 7 is the first version to use lockfile v2 which we use in this repo: https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/blob/8dfc49fc732bf6d9d9f7655213adaf7c94b532f9/package-lock.json#L4

With older npm versions, the lockfile won't work right causing builds to flake. Lockfiles are not shipped with npm releases in general, so this doesn't effect users.

aabmass:
Fixes #364

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -9,12 +9,23 @@ jobs:
     strategy:
       matrix:
         node: [10, 12, 14, 16]
+        include:
+          # use latest npm by default
+          - npm-version: latest
+          # overrides for node 10 and 12
+          - node: 10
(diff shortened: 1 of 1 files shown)


punya (approved):
LGTM but I'd prefer that we get rid of EOL versions altogether.

PR #451, opened 2022-11-22
fix: set user agent in exporters

aabmass:


packages/opentelemetry-cloud-monitoring-exporter/src/monitoring.ts
```diff
@@ -27,18 +31,29 @@ import {partitionList} from './utils';
...
+    version: VERSION,
+  },
+];
 const OT_REQUEST_HEADER = {
   'x-opentelemetry-outgoing-request': 0x1,
 };
-google.options({headers: OT_REQUEST_HEADER});
+google.options({
+  headers: OT_REQUEST_HEADER,
+  userAgentDirectives: OT_USER_AGENTS,
```
aabmass, line 55: This yields a slightly different format than we use in other exporters (see [above](https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/pull/451/files#diff-eff067595506c23393cabc321341b3a854a603444c67a52eff82e593be38264d)), b/c `googleapis-common` accepts structured user agent "directive".

I can override the WHOLE header but then we lose the `google-api-nodejs-client/5.1.0 (gzip)` part. Does anyone prefer this approach?

punya (approved):

punya (approved):

PR #455, opened 2022-11-30
fix(monitoring-exporter)!: use workload.googleapis.com as default metric prefix

aabmass:
Also removed the display name prefix. The OTel metric name without the
GCM metric prefix is now used for display name.

Fixes https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/issues/386


dashpole (approved):

aabmass:
Merging as is bc codecov isn't highlighting any code that is untested.

PR #467, opened 2022-12-10
feat(resource-util): detect GKE and GCE resource attributes

aabmass:
Implements the resource detection methods for GKE and GCE, copied from https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/tree/v1.8.0/detectors/gcp.

Added a Detector which implements the resource detection interface to glue everything together. I will add the other detectors in separate PRs.

Verified this is working with e2e test in next PR https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/pull/468

### DIFF
--- a/packages/opentelemetry-resource-util/src/detector/gke.ts
+++ b/packages/opentelemetry-resource-util/src/detector/gke.ts
@@ -1,0 +1,78 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * Implementation in this file copied from
+ * https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/v1.8.0/detectors/gcp/gke.go
+ */
+
+import * as metadata from 'gcp-metadata';
+import * as gce from './gce';
+
+const KUBERNETES_SERVICE_HOST_ENV = 'KUBERNETES_SERVICE_HOST';
+const CLUSTER_NAME_METADATA_ATTR = 'attributes/cluster-name';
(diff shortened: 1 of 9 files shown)


packages/opentelemetry-resource-util/src/detector/gke.ts
```diff
@@ -0,0 +1,78 @@
...
+ */
+
+import * as metadata from 'gcp-metadata';
+import * as gce from './gce';
+
+const KUBERNETES_SERVICE_HOST_ENV = 'KUBERNETES_SERVICE_HOST';
+const CLUSTER_NAME_METADATA_ATTR = 'attributes/cluster-name';
+const CLUSTER_LOCATION_METADATA_ATTR = 'attributes/cluster-location';
+
+export async function onGke(): Promise<boolean> {
```
dashpole, line 27: Note that this will return true on any kubernetes cluster.  The go implementation checks to see if we are on GCP first, and returns immediately if that doesn't return true:  https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/detectors/gcp/detector.go#L46

dashpole: That check would need to be used for all platforms, since it is checking to see if the metadata server is available.  Instead, the go implementation just does it once during detection

aabmass: > That check would need to be used for all platforms, since it is checking to see if the metadata server is available. Instead, the go implementation just does it once during detection

Fyi this is cached https://github.com/googleapis/google-cloud-go/blob/compute/metadata/v0.2.2/compute/metadata/metadata.go#L115-L116

Anyway, it looks like `OnGCE()` chiefly [checks for availability of the metadata server](https://github.com/googleapis/google-cloud-go/blob/ce0c9440634b17558adea724ea4e5a735dc1f1d0/compute/metadata/metadata.go#L114) along with `/sys/class/dmi/id/product_name` file.

The JS impl does not match this exactly. It has:
- [`metadata.isAvailable()`](gcpResidencyCache) check which checks for availability of metadata server.
- [`metadata.detectGcpResidency()`](https://github.com/googleapis/gcp-metadata/blob/8c0ff3293887ba4194f975405ee1c0a6e8d9dbf7/src/gcp-residency.ts#L111-L116) which does even more checks against `/sys/class/dmi/id` and for GCE MAC address.

I'm thinking to do `await isAvailable() || await detectGcpResidency()`, does that sound good?

dashpole: That sounds reasonable.  I just want to make sure this check doesn't pass outside of GCP (e.g. on AWS).

dashpole (approved):

PR #477, opened 2022-12-19
chore!: drop support for node 10 and 12

### LINKED ISSUE #476: Update supported node versions
Our stance from the README:

> The packages in this repository support all [officially supported Node.js versions](https://nodejs.org/en/about/releases/)

From the link, that is currently 14, 16, 18, 19.

aabmass:
Fixes #476

Question for the group, should we consider this a breaking change and do major version bumps on stable packages?

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -8,15 +8,10 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node: [10, 12, 14, 16]
+        node: [14, 16, 18, 19]
         include:
           # use latest npm by default
           - npm-version: latest
-          # overrides for node 10 and 12
-          - node: 10
-            npm-version: ^7
-          - node: 12
-            npm-version: ^8
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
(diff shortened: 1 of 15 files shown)


dashpole:
I think the answer is yes.  There are a bunch of open issues in the semvar repo about it with people being confused (are node engine versions a "dependency", or an API contract?), but the overall gist seems to be that a change which might requires user action is breaking.

aabmass:
That makes sense. Do we plan to bump the Go API/SDK and our exporters each time we increase the version though?

dashpole:
> That makes sense. Do we plan to bump the Go API/SDK and our exporters each time we increase the version though?

Yes, but only for major version bumps of Go.  Is this change considered a major version bump of nodejs, or a minor version bump?  I had assumed it was a major bump.

aabmass:
I guess it is major but new major versions of Node come out frequently, every ~6 month. Here is the lifecycle of current node versions for reference https://nodejs.dev/en/about/releases/
image

I dug into upstream OTel and it looks like they made a similar change in https://github.com/open-telemetry/opentelemetry-js/pull/3048 and did not bump any major versions.

aabmass:
This seems to be a frequent hot topic for semver https://github.com/semver/semver/issues/716 with no answer. I took a look at some other Google API PRs and found they consider it a breaking change, e.g. https://github.com/googleapis/nodejs-storage/pull/1876.

Let's go with that @dashpole. I've already marked this PR with `!` in the title so should be good to go :)

dashpole (approved):

PR #478, opened 2022-12-20
chore: upgrade lerna to 6.2.0

aabmass:
- chore!: drop support for node 10 and 12
- chore: upgrade lerna to 6.2.0


aabmass:
Already took care of this in another PR.

PR #480, opened 2022-12-21
chore: run e2e tests on Cloud Run

aabmass:


e2e-test-server/src/index.ts
```diff
@@ -102,13 +61,124 @@ async function pubSubPull(): Promise<void> {
...
+function pubSubPush() {
+  if (constants.PUSH_PORT === undefined) {
+    throw new Error('PUSH_PORT must be set when in push subscription mode');
+  }
+
+  const app = express();
+
+  app
+    .use(express.json())
+    .post('/', (req, res) => {
```
dashpole, line 77: do we need to handle /ready and /alive as well?

aabmass: Didn't know those are needed, what's it for?

dashpole: Ah, might just be for app-engine.

dashpole: We can add it for that later if needed.

dashpole (approved):

e2e-test-server/src/index.ts
aabmass: Ok, can add those when I add App Engine 👍

PR #482, opened 2022-12-23
feat(resource-util): add support to detect Cloud Run and Cloud Functions

aabmass:
Copied from the Go impl.

E2E tests are also passing in follow up PR https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/pull/480

### DIFF
--- a/packages/opentelemetry-resource-util/src/index.ts
+++ b/packages/opentelemetry-resource-util/src/index.ts
@@ -21,8 +21,12 @@ import {
 
 const AWS_ACCOUNT = 'aws_account';
 const AWS_EC2_INSTANCE = 'aws_ec2_instance';
+const CLOUD_FUNCTION = 'cloud_function';
+const CLOUD_RUN_REVISION = 'cloud_run_revision';
 const CLUSTER_NAME = 'cluster_name';
+const CONFIGURATION_NAME = 'configuration_name';
 const CONTAINER_NAME = 'container_name';
+const FUNCTION_NAME = 'function_name';
 const GCE_INSTANCE = 'gce_instance';
 const GENERIC_NODE = 'generic_node';
 const GENERIC_TASK = 'generic_task';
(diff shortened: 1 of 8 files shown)


packages/opentelemetry-resource-util/src/index.ts
```diff
@@ -110,6 +116,16 @@ const MAPPINGS = {
...
   },
+  [CLOUD_RUN_REVISION]: {
+    [LOCATION]: {otelKeys: [SemanticResourceAttributes.CLOUD_REGION]},
+    [SERVICE_NAME]: {otelKeys: [SemanticResourceAttributes.FAAS_NAME]},
+    [CONFIGURATION_NAME]: {otelKeys: [SemanticResourceAttributes.FAAS_NAME]},
+    [REVISION_NAME]: {otelKeys: [SemanticResourceAttributes.FAAS_VERSION]},
+  },
+  [CLOUD_FUNCTION]: {
+    [REGION]: {otelKeys: [SemanticResourceAttributes.CLOUD_REGION]},
+    [FUNCTION_NAME]: {otelKeys: [SemanticResourceAttributes.FAAS_NAME]},
```
aabmass, line 127: These mappings are used for both trace `g.co/r/` labels and monitoring monitored resource. These monitored resources are not currently writable for monitoring, so this presents an issue (we have the same issue in the java impl I believe https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/blob/60cf3d675eda0900258853d568c21bfd4d692d49/shared/resourcemapping/src/main/java/com/google/cloud/opentelemetry/resource/ResourceTranslator.java#L97

Should we go ahead with this or maybe I can add a `skipUnsupportedMonitoredResources` bool option?

dashpole: You can see what I did in go: https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/main/exporter/metric/metric.go#L343

Either option is fine with me

aabmass: Disabled auto-merge. I will update it so it works correctly on cloud run

aabmass: Updated L187 ptal

dashpole (approved):

packages/opentelemetry-resource-util/src/index.ts
mbrevda: appologies for waking a dead thread: could you possible explain "These monitored resources are not currently writable for monitoring" and/or point to reference documentation?

PR #486, opened 2022-12-29
feat: run e2e tests on Cloud Functions

aabmass:
One small tweak in ordering of resource checks was needed to make this work. Probably need the same change in opentelemetry-operations-go.

Otherwise, got monorepo working with Cloud Functions by packing the dependency sources into tarballs (with `npm pack`) and including that in the zip source.


dashpole (approved):

dashpole:
Yep.  I had to make the same change in go: https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/pull/552/commits/792adca74f8dec01faa63e3d3428badb162f30a9

aabmass:
@dashpole going to ship this for now and then address where we should be doing `g.co/r/cloud_function/...` span labels.

PR #498, opened 2023-01-09
test: update Dockerfile to not use lerna for builds

aabmass:


aabmass:
This is too complicated and I still can't figure out how to get it working correctly. It seems like docker sometimes won't re-use the cache and idk why.

PR #506, opened 2023-01-13
test: run e2e tests on gae

aabmass:


aabmass:
Squashed into the same PR as the detector

PR #536, opened 2023-03-17
chore: have renovate update snapshot tests after upgrading OTel versions

aabmass:


punya (approved):
Cool, I didn't know you could make Renovate do extra stuff after an upgrade.

PR #560, opened 2023-06-24
feat(monitoring): implement ExponentialHistogram exporting

### LINKED ISSUE #555: Support exponential histogram in the exporter
This was added in the 1.14 and needs to be wired into the exporter.

aabmass:
Fixes #555

Copied the implementation from the collector exporter and added tests. Note this is not super useful until https://github.com/open-telemetry/opentelemetry-js/issues/3931 is fixed and released.


dashpole (approved):

justin--:
Will a new version be published that includes this and other latest changes?

I also see that https://github.com/open-telemetry/opentelemetry-js/issues/3931 was closed and looks to have a fix in. Does that also need a version bump? 

Appreciate getting this addressed though, excited to be able to use this!

PR #563, opened 2023-06-26
feat(resource-util): implement the DetectorSync resource detector interface

aabmass:
For more context, this new interface was added upstream (and the old one deprecated) so that resource detection will be "synchronous" and the SDK initialization can complete before any other code runs. For auto instrumentation, this is important so that monkey patching is finished before the user imports anything.

The deprecated `GcpDetector` will be removed in the next major version

### DIFF
--- a/packages/opentelemetry-resource-util/src/detector/detector.ts
+++ b/packages/opentelemetry-resource-util/src/detector/detector.ts
@@ -26,158 +26,150 @@ import * as gae from './gae';
 import * as metadata from 'gcp-metadata';
 import {Attributes} from '@opentelemetry/api';
 
-export class GcpDetectorBase {
-  // Cached project ID
-  private _project: string | undefined;
-
-  protected async _detect(): Promise<Resource> {
-    if (!(await metadata.isAvailable())) {
-      return Resource.EMPTY;
-    }
-
-    // Note the order of these if checks is significant with more specific resources coming
-    // first. E.g. Cloud Functions gen2 are executed in Cloud Run so it must be checked first.
-    if (await gke.onGke()) {
-      return await this._gkeResource();
-    } else if (await faas.onCloudFunctions()) {
-      return await this._cloudFunctionsResource();
-    } else if (await faas.onCloudRun()) {
-      return await this._cloudRunResource();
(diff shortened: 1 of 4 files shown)


packages/opentelemetry-resource-util/test/detector/detector.test.ts
```diff
@@ -207,3 +207,202 @@ describe('GcpDetector', () => {
     assert.deepStrictEqual(resource.attributes, {});
   });
 });
+
+describe('GcpDetectorSync', () => {
```
aabmass, line 211: These tests were copy pasted from above and I only updated

```diff
-const resource = await new GcpDetector().detect();
+const resource = new GcpDetectorSync().detect();
+await resource.waitForAsyncAttributes?.();
```

packages/opentelemetry-resource-util/src/detector/detector.ts
```diff
@@ -18,19 +18,19 @@ import {
...
+import {Detector, DetectorSync, Resource} from '@opentelemetry/resources';
 import * as gce from './gce';
 import * as gke from './gke';
 import * as faas from './faas';
 import * as gae from './gae';
 import * as metadata from 'gcp-metadata';
 import {Attributes} from '@opentelemetry/api';
 
-export class GcpDetector implements Detector {
+export class GcpDetectorBase {
```
punya, line 29: Nit: I generally avoid using class inheritance to model shared implementation. How about leaving `GcpDetector` unchanged (with the exception of a deprecation comment), and writing `GcpDetectorSync` using composition:

```ts
export class GcpDetectorSync implements DetectorSync {
  private asyncDetector = new GcpDetector();

  detect(): Resource {
    return new Resource({}, asyncDetector().then(res => res.attributes));
  }
}
```

punya (approved):

packages/opentelemetry-resource-util/src/detector/detector.ts
aabmass: I refactored the impl into a standalone function. The class was only being used for caching `projectId` which idek why I added, as it never gets re-read in the one-shot case.

PR #612, opened 2023-08-24
Use provider.register instead of setGlobalTraceProvider

hongalex:
Calling `provider.register()` calls `setGlobalTracerProvider` and also does some additional tweaking that makes context propagation work properly. Even though this sample doesn't do context propagation, this would save developers headache when they're copying and pasting the setup from this example and wondering why their propagators aren't working with instrumented libraries (which is what happened with me).


aabmass (approved):
Thanks for the PR!

aabmass:
@hongalex can you fix the commit message to follow convention commits?

aabmass:
:( not sure why it auto merged since that was a required check, oops

PR #622, opened 2023-09-19
Issue 620: Timestamps with floating point numbers cause metric processing to get stuck

### LINKED ISSUE #620: Timestamps with decimals cause metric forwarding to get stuck
    "@opentelemetry/api": "1.4.1",
    "@opentelemetry/core": "1.15.2",
    "@opentelemetry/resources": "1.15.2",
    "@opentelemetry/sdk-metrics": "1.15.2",

v16.14.2

### What did you do?
Very occasionally, metrics created with OpenTelemetry end up with a decimal part to their timestamps. This is presumably due to numeric precision issues causing integer math to create a floating point number. Attempting to process metrics with floating point timestamps causes a RangeError exception to be thrown. The exception isn't caught until somewhere well up the stack. A side effect is that the list of pending metrics is not cleared out, resulting in this error being hit every time the system tries to forward the list of pending metrics. The following is an example of a metric seen when this problem was encountered:

```{
    "attributes":
    {
    },
    "startTime":
    [
        1695096840,
        534000118.00000006
    ],
    "endTime":
    [
        1695096846,
        89824450
    ],
    "value": 22
```
The function that transforms the metrics should:
1. Round timestamps to prevent inadvertent decimal values from creating problems
2. Catch any exceptions to prevent them from bubbling up the stack and bypassing clean-up code. There is already a precedent in the function for discarding invalid metrics. This would extend it to discard any metrics where an exception was thrown while it was being processed.

I'll submit a PR shortly with the above fixes.

sethrwebster:
This PR does two things:
1. Rounds timestamp values to ensure they are integers, as floating point numbers will break downstream processing
2. Wraps a try/catch around the metric processing to make sure any other errors don't cause the invalid metric to be left in place, blocking all future metric processing.

A majority of the changes in this PR are whitespace changes from wrapping a try/catch around a block of code.

Fixes #620


packages/opentelemetry-cloud-monitoring-exporter/test/transform.test.ts
```diff
@@ -27,6 +27,44 @@ const {normalizeLabelKey} = _TEST_ONLY;
 
 describe('transform', () => {
   let diagSpy: sinon.SinonSpiedInstance<DiagAPI>;
+  const expected = [
```
sethrwebster, line 30: Moved this outside of the tests because it is now common to two tests.

packages/opentelemetry-cloud-monitoring-exporter/test/transform.test.ts
```diff
@@ -115,51 +153,58 @@ describe('transform', () => {
...
         ],
       };
-
       assert.deepStrictEqual(
         createTimeSeries(
           data,
           {labels: {}, type: ''},
           'workload.googleapis.com'
         ),
-        [
```
sethrwebster, line 125: Diff butchered this change. This test now contain a reference to the expected data structure above. The rest of the file is a new test.

sethrwebster:
Closing this PR as the problem was traced to an old version of `opentelemetry-js` being used. The issue of creating invalid timestamps has been fixed in version 1.9.1.

PR #638, opened 2023-10-05
fix(resource-detector): map metadata server's instanceID to faas.id instead of faas.instance_id

aabmass:
This matches the current Go/collector impl:
- [`faas.go` in operations-go](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/6b51e22bb5a89706db8e631469b1a491a0a693c3/detectors/gcp/faas.go#L90-L93)
- [`gcp.go` in collector-contrib](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/21ee863aaf7e4d19821927d5108fdf7dd720ec0c/processor/resourcedetectionprocessor/internal/gcp/gcp.go#L83)

This looks like a bug copied from the original Go impl https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/4e508ee3bbb979320283005464642e7d1afd0842/detectors/gcp/faas.go#L59-L62

### DIFF
--- a/packages/opentelemetry-resource-util/src/detector/detector.ts
+++ b/packages/opentelemetry-resource-util/src/detector/detector.ts
@@ -66,37 +66,35 @@ async function gkeResource(): Promise<Resource> {
 }
 
 async function cloudRunResource(): Promise<Resource> {
-  const [faasName, faasVersion, faasInstance, faasCloudRegion] =
-    await Promise.all([
-      faas.faasName(),
-      faas.faasVersion(),
-      faas.faasInstance(),
-      faas.faasCloudRegion(),
-    ]);
+  const [faasName, faasVersion, faasId, faasCloudRegion] = await Promise.all([
+    faas.faasName(),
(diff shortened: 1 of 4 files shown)


aabmass:
Ok now this causes tests to fail because the e2e test runner is expecting `faas.instance` after all: https://github.com/GoogleCloudPlatform/opentelemetry-operations-e2e-testing/blob/d72a3b50c2e9506009c89a831b435c795edcb9c8/trace_test.go#L258

And that is what we send in SDK detector: https://github.com/open-telemetry/opentelemetry-go-contrib/blob/8a2f8cfc4ac7dd7b9c1771cfbff2676f7777652e/detectors/gcp/detector.go#L63C25-L63C25

Will open a separate issue to discuss

aabmass:
Oops I'm just confused https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/issues/679

PR #678, opened 2024-02-29
Add OTLP trace http sample with Google auth

damemi:
To run:

```
export OTEL_RESOURCE_ATTRIBUTES="gcp.project_id=<project-id>"
export OTEL_EXPORTER_OTLP_ENDPOINT=<endpoint>
node app.js
```

In another terminal
```
curl localhost:8080/rolldice?rolls=12
```
(or any number of rolls)

Samples heavily based on OTel docs at https://opentelemetry.io/docs/languages/js/getting-started/nodejs, so adding their copyright as well

### DIFF
--- a/samples/otlptracehttp/package.json
+++ b/samples/otlptracehttp/package.json
@@ -1,0 +1,23 @@
+{
+  "name": "sample-app",
+  "version": "0.1.0",
+  "private": true,
+  "license": "Apache-2.0",
+  "author": "Google Inc.",
+  "main": "index.js",
+  "scripts": {
+    "start": "node ./app.js"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
+  "dependencies": {
+    "@opentelemetry/api": "1.7.0",
+    "@opentelemetry/exporter-trace-otlp-http": "0.49.1",
+    "@opentelemetry/resources": "1.19.0",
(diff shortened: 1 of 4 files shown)


samples/otlptracehttp/package.json
```diff
@@ -0,0 +1,24 @@
...
+  "license": "ISC",
+  "dependencies": {
+    "@opentelemetry/api": "^1.8.0",
+    "@opentelemetry/exporter-metrics-otlp-http": "^0.49.1",
+    "@opentelemetry/exporter-trace-otlp-http": "^0.49.1",
+    "@opentelemetry/resources": "^1.22.0",
+    "@opentelemetry/sdk-node": "^0.49.1",
+    "@opentelemetry/semantic-conventions": "^1.22.0",
+    "express": "^4.18.3",
+    "google-auth-library": "^9.6.3"
```
aabmass, line 22: Since this is an example, can you remove the carats so that it installs exact versions?

One other thing, this repo uses npm workspaces where there is a single package-lock.json in the repo root and all of the packages get linked together (similar to go workspaces). Can you delete the package-lock.json and `node_modules` from this directory and instead run `npm install` from the repo root?

damemi: Done, please check if I missed anything

aabmass (approved):

PR #687, opened 2024-03-21
docs: OTel quickstart showing auto instrumentation of an express app

aabmass:


samples/instrumentation-quickstart/README.md
```diff
@@ -0,0 +1,67 @@
...
+```
+
+### Locally with Application Default Credentials
+
+First Create local credentials by running the following command and following the
+oauth2 flow (read more about the command [here][auth_command]):
+
+	gcloud auth application-default login
+
+Set the `GOOGLE_APPLICATION_CREDENTIALS` environment variable with `export GOOGLE_APPLICATION_CREDENTIALS="$HOME/.config/gcloud/application_default_credentials.json"`
```
dashpole, line 46: Non-blocking: Isn't this already the default for ADC?  Do users actually need to set this?

samples/instrumentation-quickstart/docker-compose.creds.yaml
```diff
@@ -0,0 +1,35 @@
+# Copyright 2023 Google LLC
```
dashpole, line 1:
```suggestion
# Copyright 2024 Google LLC
```

samples/instrumentation-quickstart/docker-compose.yaml
```diff
@@ -0,0 +1,49 @@
+# Copyright 2023 Google LLC
```
dashpole, line 1:
```suggestion
# Copyright 2024 Google LLC
```

samples/instrumentation-quickstart/otel-collector-config.yaml
```diff
@@ -0,0 +1,111 @@
+# Copyright 2023 Google LLC
```
dashpole, line 1:
```suggestion
# Copyright 2024 Google LLC
```

samples/instrumentation-quickstart/src/index.ts
```diff
@@ -0,0 +1,69 @@
...
+
+// [START opentelemetry_instrumentation_handle_single]
+/**
+ * handleSingle handles an http request by sleeping for 100-200 ms. It writes the number of
+ * milliseconds slept as its response.
+ */
+fastify.get('/single', async request => {
+  // Sleep between 100-200 milliseconds
+  const sleepMillis = randInt(100, 200);
+  request.log.info({sleepMillis}, 'Going to sleep');
```
dashpole, line 36: Non-blocking: Neat syntax.  Does this end up structured?

dashpole (commented):
Does this set the http.route?

samples/instrumentation-quickstart/README.md
aabmass: `docker-compose.creds.yaml` uses the environment variable as the source for the volume mount of credentials into the collector container. The compose file could default the environment variable to this, but the intention is to make it clear to the user that the sample will mount their creds. The path would also be different on Windows.

Do you think it would be better to default it?

samples/instrumentation-quickstart/docker-compose.yaml
aabmass: These files were technically copied from another repo but Im fine to update them

samples/instrumentation-quickstart/src/index.ts
aabmass: Yup!
image

aabmass (commented):
Yes it has `http.route`. It's in metrics too, see the screenshot above.
image

samples/instrumentation-quickstart/README.md
dashpole: Got it.  So if we defaulted, the sample wouldn't work on windows?  If that is the case, the current implementation is good.

aabmass: Windows users would need to override the default. I haven't actually tried it on Windows though.

In case it wasn't clear, if they run on GCE or Cloud Shell they don't need to set any environment variables, just `docker compose up`

dashpole (approved):

PR #704, opened 2024-04-27
feat: add trace export sample via http/protobuf

psx95:
Very similar to existing samples, except that the OTLP exporter used is the one for http with protobuf.

### Testing
 - `npm run lint`
 - `npm run start-http-proto`

The existing client can be re-used for indefinite export.


aabmass (commented):
Are we going to fold the https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/tree/main/samples/otlptracehttp sample in here as well or did you want to do a follow up PR?

samples/otlptraceexport/src/app-http-proto-export.ts
```diff
@@ -0,0 +1,60 @@
...
+
+import {NodeSDK} from '@opentelemetry/sdk-node';
+import {OTLPTraceExporter} from '@opentelemetry/exporter-trace-otlp-proto';
+import {AuthClient, GoogleAuth} from 'google-auth-library';
+
+const PORT = parseInt(process.env.PORT || '8080');
+const app = express();
+
+async function getAuthenticatedClient(): Promise<AuthClient> {
+  const auth: GoogleAuth = new GoogleAuth({
```
aabmass, line 26: Seems to have been last PR
```suggestion
  const auth = new GoogleAuth({
```

samples/otlptraceexport/src/app-http-proto-export.ts
```diff
@@ -0,0 +1,60 @@
...
+    scopes: 'https://www.googleapis.com/auth/cloud-platform',
+  });
+  return await auth.getClient();
+}
+
+// Express App that exports traces over HTTP with protobuf
+async function main() {
+  const authenticatedClient: AuthClient = await getAuthenticatedClient();
```
aabmass, line 34: Any reason to explicitly annotate here?

aabmass (approved):

damemi (approved):

psx95:
> Are we going to fold the https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/tree/main/samples/otlptracehttp sample in here as well or did you want to do a follow up PR?

Was planning to do that in a follow-up PR

samples/otlptraceexport/src/app-http-proto-export.ts
psx95: Nothing other than making it explicit for the reader. I can remove it to adhere to the Typescript style guide.

PR #705, opened 2024-04-29
refactor: move http/json export in unified traceexport folder

psx95:
Moves otlp-trace-export http/json sample in the unified sample structure.


samples/otlptraceexport/README.md
```diff
@@ -24,6 +25,7 @@ cd samples/otlptracexport && npm run start
 ```
 Alternate run scripts available:
```
aabmass, line 26: Any reason to give gRPC the shorter name of `start` instead of matching these names?

samples/otlptraceexport/src/app-http-json-export.ts
```diff
@@ -1,4 +1,3 @@
-// Copyright OpenTelemetry Authors
```
aabmass, line 1: Maybe we should keep this. I think @damemi originally added it in the first sample since it was based on the upstream quickstart.

aabmass (approved):

samples/otlptraceexport/README.md
psx95: Just that it was chosen be default. People are familiar with `npm run start` - so just wanted that command to work.

samples/otlptraceexport/src/app-http-json-export.ts
psx95: The sample has changed and is extremely similar to other scripts which do not have this copyright. I don't mind keeping it, but I would defer this it @damemi.

damemi: Yeah when I added this in https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/pull/678, a lot of the code was copied directly from that page. Other language samples that I wrote had enough existing code in our repos that I didn't need to copy and paste so much from upstream for them, so they don't have the upstream copyright.

But yeah, this looks like the code has changed a lot from what I added. I don't know what the exact rule is on removing the copyright header, but this is basically a different file now, so I don't think we still need it.

psx95: Thanks for the info, @damemi.

I think then I will not add it back then.

Issue #655, opened 2023-10-24
Cloud monitoring export reports "invalid location or region" in GAE nodejs20 


fpoppinga:
Please answer these questions before submitting a bug report.

### What version of OpenTelemetry are you using?

Excerpt from my `package.json`: 
```json
    "@google-cloud/opentelemetry-cloud-monitoring-exporter": "^0.17.0",
    "@google-cloud/opentelemetry-cloud-trace-exporter": "^2.1.0",
    "@google-cloud/opentelemetry-resource-util": "^2.1.0",
    "@opentelemetry/api": "^1.6.0",
    "@opentelemetry/core": "^1.17.1",
    "@opentelemetry/instrumentation": "^0.44.0",
    "@opentelemetry/instrumentation-express": "^0.33.2",
    "@opentelemetry/instrumentation-http": "^0.44.0",
    "@opentelemetry/instrumentation-pg": "^0.36.2",
    "@opentelemetry/resources": "^1.17.1",
    "@opentelemetry/sdk-metrics": "^1.17.1",
    "@opentelemetry/sdk-trace-base": "^1.17.1",
    "@opentelemetry/sdk-trace-node": "^1.17.1",
    "@opentelemetry/semantic-conventions": "^1.17.1",
```

### What version of Node are you using?
The app is running on the GAE app engine [standard environment](https://cloud.google.com/appengine/docs/standard/nodejs/runtime) `nodejs20`.  

### What did you do?

- Setup the cloud monitoring according to the documentation in the README.md
- observe, no metrics are reported when running in GAE standard environment
- enable `diag` logging 
- observe the error message below

### What did you expect to see?

Metrics being reported to cloud monitoring. 

### What did you see instead?

```
PeriodicExportingMetricReader: metrics export failed (error Error: Send TimeSeries failed: One or more TimeSeries could not be written: Unrecognized region or location.: gae_instance{instance_id:0037d<redacted>da5b0,version_id:20231024t093347,module_id:default,location:projects/1<redacted>5/zones/eu2} timeSeries[0]: workload.googleapis.com/order.erp_order_change_upsert{orgId:hmd})`
```

### Additional context

- Reporting metrics worked fine with a prehistoric version of opentelemetry before, but in an effort to upgrade all dependencies, metric reporting stopped working.
- I assume that this might have to do with the resource detection. We're using the `import { GcpDetectorSync } from "@google-cloud/opentelemetry-resource-util";`.

[...]

PR #715, opened 2024-05-03
fix(resource-detector): properly detect zone on AppEngine standard

aabmass:
Fixes https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/issues/655

It turns out GAE standard environment's metadata server returns a zone string like `projects/myproject/zones/us15`. ut integration tests only run on AppEngine flex not standard so this wasn't caught 🙁. Here are HTTP logs I printed out from a sample app:

```json
{
  "jsonPayload": {
    "message": "gaxios metadata request",
    "responseData": "projects/<project>/zones/us15",
    "requestUrl": "http://169.254.169.254/computeMetadata/v1/instance/zone"
  },
  "resource": {
    "type": "gae_app",
    "labels": {
      "zone": "us15",
      "version_id": "20240503t222051",
      "project_id": "otel-starter-project",
      "module_id": "default"
    }
  }
}
```

I updated the tests and manually tested a sample app on GAE standard writing metrics: 
image

---
I think this bug is also affecting all other resource detectors e.g. [Go's impl that this was copied from](https://github.com/GoogleCloudPlatform/opentelemetry-operations-go/blob/bda7f022e62f4ec2c0c112a0f4485b8c89746bcd/detectors/gcp/app_engine.go#L68-L71)

### DIFF
--- a/packages/opentelemetry-resource-util/src/detector/detector.ts
+++ b/packages/opentelemetry-resource-util/src/detector/detector.ts
@@ -123,10 +123,8 @@ async function gaeResource(): Promise<Resource> {
       gae.standardAvailabilityZone(),
       gae.standardCloudRegion(),
     ]);
-    console.log('Got GAE standard zone=%s and region=%s', zone, region);
   } else {
     ({zone, region} = await gce.availabilityZoneAndRegion());
-    console.log('Got GAE flex zone=%s and region=%s', zone, region);
   }
   const [faasName, faasVersion, faasInstance] = await Promise.all([
     gae.serviceName(),
--- a/packages/opentelemetry-resource-util/src/detector/detector.ts
+++ b/packages/opentelemetry-resource-util/src/detector/detector.ts
@@ -123,8 +123,10 @@ async function gaeResource(): Promise<Resource> {
       gae.standardAvailabilityZone(),
       gae.standardCloudRegion(),
     ]);
+    console.log('Got GAE standard zone=%s and region=%s', zone, region);
   } else {
     ({zone, region} = await gce.availabilityZoneAndRegion());
+    console.log('Got GAE flex zone=%s and region=%s', zone, region);
   }
   const [faasName, faasVersion, faasInstance] = await Promise.all([
     gae.serviceName(),
--- a/packages/opentelemetry-resource-util/src/detector/gae.ts
+++ b/packages/opentelemetry-resource-util/src/detector/gae.ts
@@ -76,7 +76,9 @@ export async function flexAvailabilityZoneAndRegion(): Promise<{
  * true before calling this, or it may throw exceptions.
  */
 export async function standardAvailabilityZone(): Promise<string> {
-  return await metadata.instance<string>(ZONE_METADATA_ATTR);
+  const zone = await metadata.instance<string>(ZONE_METADATA_ATTR);
+  // zone is of the form "projects/233510669999/zones/us15"
+  return zone.slice(zone.lastIndexOf('/') + 1);
 }
 
 /**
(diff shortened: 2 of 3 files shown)


packages/opentelemetry-resource-util/src/detector/detector.ts
```diff
@@ -123,8 +123,10 @@ async function gaeResource(): Promise<Resource> {
       gae.standardAvailabilityZone(),
       gae.standardCloudRegion(),
     ]);
+    console.log('Got GAE standard zone=%s and region=%s', zone, region);
```
psx95, line 126: These were probably debug statements.

psx95:
- The [Java Resource detector implementation](https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/blob/da9f22582113d93e2b23cc4ffaf42f294b03fb6b/detectors/resources-support/src/main/java/com/google/cloud/opentelemetry/detection/GoogleAppEngine.java#L44) does not seem to have this issue.

- The [Python resource detector](https://github.com/GoogleCloudPlatform/opentelemetry-operations-python/tree/main/opentelemetry-resourcedetector-gcp/src/opentelemetry/resourcedetector/gcp_resource_detector) does not seem to support GAE. Is this correct ?

psx95 (changes requested):
Nit: The `console.log` statements could be removed.

The change looks good.

packages/opentelemetry-resource-util/src/detector/detector.ts
aabmass: Oops thanks

aabmass:
> * The [Java Resource detector implementation](https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/blob/da9f22582113d93e2b23cc4ffaf42f294b03fb6b/detectors/resources-support/src/main/java/com/google/cloud/opentelemetry/detection/GoogleAppEngine.java#L44) does not seem to have this issue.

I think it does have the issue. [This test](https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/blob/da9f22582113d93e2b23cc4ffaf42f294b03fb6b/detectors/resources-support/src/test/java/com/google/cloud/opentelemetry/detection/GCPPlatformDetectorTest.java#L319-L320) doesn't provide the slash separated string for GAE standard and the detector just returns the zone directly

dashpole (approved):

psx95:
> > * The [Java Resource detector implementation](https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/blob/da9f22582113d93e2b23cc4ffaf42f294b03fb6b/detectors/resources-support/src/main/java/com/google/cloud/opentelemetry/detection/GoogleAppEngine.java#L44) does not seem to have this issue.
> 
> I think it does have the issue. [This test](https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/blob/da9f22582113d93e2b23cc4ffaf42f294b03fb6b/detectors/resources-support/src/test/java/com/google/cloud/opentelemetry/detection/GCPPlatformDetectorTest.java#L319-L320) doesn't provide the slash separated string for GAE standard and the detector just returns the zone directly

We might need to add a test to prevent regression, but going by the [implementation of `metadataConfig.getZone()`](https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/blob/da9f22582113d93e2b23cc4ffaf42f294b03fb6b/detectors/resources-support/src/main/java/com/google/cloud/opentelemetry/detection/GCPMetadataConfig.java#L67), it does not look like there would be an issue as it is similar to what is being done in this fix.

psx95 (approved):

aabmass:
> We might need to add a test to prevent regression, but going by the [implementation of `metadataConfig.getZone()`](https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/blob/da9f22582113d93e2b23cc4ffaf42f294b03fb6b/detectors/resources-support/src/main/java/com/google/cloud/opentelemetry/detection/GCPMetadataConfig.java#L67), it does not look like there would be an issue as it is similar to what is being done in this fix.

Here is a test update https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/pull/341 thanks

PR #717, opened 2024-05-07
test: run e2e tests on GAE standard

aabmass:
The yaml is mostly copied from the Cloud Functions config which also outputs a zip file.

### DIFF
--- a/cloudbuild-e2e-gae-standard.yaml
+++ b/cloudbuild-e2e-gae-standard.yaml
@@ -1,4 +1,4 @@
-# Copyright 2021 Google LLC
+# Copyright 2024 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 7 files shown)


cloudbuild-e2e-gae-standard.yaml
```diff
@@ -0,0 +1,53 @@
...
+
+      # create a new fake package.json in dist/ and zip it up into a source zip for GAE
+      # standard
+      cd dist
+      cp ../fake.package.json package.json
+      zip -qr appsource.zip .
+
+  # Run the test
+  - name: $_TEST_RUNNER_IMAGE
+    id: run-tests-cloud-functions
```
dashpole, line 41: update this?

aabmass: Good catch, fixed

dashpole (approved):

PR #718, opened 2024-05-07
refactor: CallCredetials creation process

psx95:
Minor refactor to incorporate a suggested improvement on composing composite ChannelCredentials in grpc-js. 

### Testing
 - `npm run start` && `npm run client` was able to successfully export traces to Google Cloud.

### DIFF
--- a/samples/otlptraceexport/src/app-grpc-export.ts
+++ b/samples/otlptraceexport/src/app-grpc-export.ts
@@ -18,7 +18,7 @@ import {rollTheDice} from './dice';
 import {NodeSDK} from '@opentelemetry/sdk-node';
 import {OTLPTraceExporter} from '@opentelemetry/exporter-trace-otlp-grpc';
 import {AuthClient, GoogleAuth} from 'google-auth-library';
-import {credentials} from '@grpc/grpc-js';
+import {CallCredentials, ChannelCredentials} from '@grpc/grpc-js';
 
 const PORT = parseInt(process.env.PORT || '8080');
 const app = express();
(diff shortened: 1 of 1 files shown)


dashpole:
I assume we can't just use https://github.com/grpc/grpc-node/blob/0ae86b13f0b3492293856be7f224fe1a4eb3432d/packages/grpc-js-xds/src/google-default-credentials.ts#L21 directly (or it isn't a dependency we want to take)?

aabmass:
> Minor refactor to incorporate a suggested improvement on composing composite ChannelCredentials in grpc-js.

Could you share a link? This looks maybe a little more complicated than the current code if they are functionally equivalent

psx95:
> I assume we can't just use https://github.com/grpc/grpc-node/blob/0ae86b13f0b3492293856be7f224fe1a4eb3432d/packages/grpc-js-xds/src/google-default-credentials.ts#L21 directly (or it isn't a dependency we want to take)?

We can use it, but we would need to add a dependency on `@grpc/grpc-js-xds`.  There's no harm in taking this dependency for this sample, but I thought sample should avoid additional dependencies if possible, as it might give a notion to the reader that the dependency is **required**, with no way around it.

aabmass:
> We can use it, but we would need to add a dependency on `@grpc/grpc-js-xds`

Reading that code, it also doesn't look like the function is re-exported from the index.ts which typically means it's not part of the public API.

samples/otlptraceexport/src/app-grpc-export.ts
```diff
@@ -33,12 +33,12 @@ async function getAuthenticatedClient(): Promise<AuthClient> {
 // Express App that exports traces via gRPC with protobuf
 async function main() {
   const authenticatedClient: AuthClient = await getAuthenticatedClient();
+  const sslCreds: ChannelCredentials = ChannelCredentials.createSsl();
 
   const sdk = new NodeSDK({
     traceExporter: new OTLPTraceExporter({
-      credentials: credentials.combineChannelCredentials(
-        credentials.createSsl(),
-        credentials.createFromGoogleCredential(authenticatedClient)
```
aabmass, line 41: IMO this is more readable and less complex, so would be my preference if they are functionally equivalent.

psx95: I can close this PR as I do not have any particular preference in this case.

PR #728, opened 2024-06-28
feat: allow setting a custom user agent and version.

kmandrika:
- [ ] Make sure to open an issue as a [bug/issue/feature](https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [ ] Ensure the tests and linter pass

Fixes #0

### DIFF
--- a/packages/opentelemetry-cloud-monitoring-exporter/test/instrument-snapshot.test.ts
+++ b/packages/opentelemetry-cloud-monitoring-exporter/test/instrument-snapshot.test.ts
@@ -355,12 +355,36 @@ describe('MetricExporter snapshot tests', () => {
       });
     });
   });
+
+  describe('Custom User Agent', () => {
+    it('correctly adds a custom user agent when one is specified', async () => {
+      const metric = await generateMetricsData((_, meter) => {
+        const counter = meter.createCounter('mycounter', {
+          description: 'description',
+          unit: 'unit',
+          valueType: ValueType.INT,
+        });
+        counter.add(1);
+      });
+
+      const result = await callExporter(metric, {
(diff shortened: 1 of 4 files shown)


aabmass (commented):
@kmandrika LGTM, can you just add a test?

The golden/snapshot tests in https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/blob/main/packages/opentelemetry-cloud-monitoring-exporter/test/instrument-snapshot.test.ts already check the user agent. You can add another case in there or use that as an example.

The convention commits check is also failing

packages/opentelemetry-cloud-monitoring-exporter/test/instrument-snapshot.test.ts
```diff
@@ -355,12 +355,36 @@ describe('MetricExporter snapshot tests', () => {
...
 
-function callExporter(resourceMetrics: ResourceMetrics): Promise<ExportResult> {
+function callExporter(
+  resourceMetrics: ResourceMetrics,
+  userAgent?: any
+): Promise<ExportResult> {
   return new Promise(resolve => {
     const exporter = new MetricExporter({
       projectId: PROJECT_ID,
+      userAgent,
```
aabmass, line 387:
```suggestion
  exporterOptions?: ExporterOptions
): Promise<ExportResult> {
  return new Promise(resolve => {
    const exporter = new MetricExporter({
      projectId: PROJECT_ID,
      ...exporterOptions,
```

aabmass (approved):
One comment otherwise looks good

PR #730, opened 2024-07-12
chore: fix tsconfigs to all be the same

aabmass:
By fixing the `extends` to be the same. Previous versions of typescript required a path but now an NPM import path works. For the e2e-test package, i had to leave as is because ncc breaks and I don't have time to look into it.

### DIFF
--- a/samples/otlptraceexport/tsconfig.json
+++ b/samples/otlptraceexport/tsconfig.json
@@ -1,8 +1,12 @@
 {
-  "extends": "../../node_modules/gts/tsconfig-google.json",
+  "extends": "gts/tsconfig-google.json",
   "compilerOptions": {
     "rootDir": ".",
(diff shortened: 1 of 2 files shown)


samples/otlptraceexport/tsconfig.json
```diff
@@ -1,8 +1,12 @@
 {
-  "extends": "../../node_modules/gts/tsconfig-google.json",
+  "extends": "gts/tsconfig-google.json",
   "compilerOptions": {
     "rootDir": ".",
     "outDir": "build",
     "esModuleInterop": true
-  }
+  },
```
aabmass, line 7: Without this, typescript tries to compile the `build/` directory and complains about overwriting input files

punya (approved):

PR #741, opened 2024-08-04
refactor: Use JSON files for TraceExporter client initialization

### LINKED ISSUE #404: Loading of protos at runtime does not place nice with bundlers

1.2.0

16

### What did you do?

I'm using the Google trace exporter in a Next.js project as well as with a backend service bundled using esbuild. In both of these cases, the trace exporter fails at runtime when trying to initializer grpc. 

```
Google Cloud Trace authenticating
Google Cloud Trace got authentication. Initializaing rpc client
failed to create client: ENOENT: no such file or directory, open '[mydir]/google/devtools/cloudtrace/v2/tracing.proto'
{"stack":"Error: failed to create client: ENOENT: no such file or directory, open '[mydir]/google/devtools/cloudtrace/v2/tracing.proto'","message":"failed to create client: ENOENT: no such file or directory, open '[mydir]/google/devtools/cloudtrace/v2/tracing.proto'","errno":"-2","code":"ENOENT","syscall":"open","path":"[mydir]/google/devtools/cloudtrace/v2/tracing.proto","name":"Error"}
```

"[mydir]" is part of my project directory structure, but in fact one level above the project root when running from `dist` and a bundled version of my service.

It appears that the shared google proto library assumes that it lives in a flat `node_modules` directory and tries to find the protos in `../..` relative to it - https://github.com/googleapis/nodejs-proto-files/blob/main/src/index.ts#L19, referenced from https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/blob/main/packages/opentelemetry-cloud-trace-exporter/src/trace.ts#L141

A similar error occurs when running under Next.

Protos being found automatically.

### What did you see instead?

A proto file failed to load due to assumptions about directory structure.

### Additional context

Ideally the protos would be bundled in such a way that esbuild or other tools could include the content, e.g. as a string content instead of a .proto file, but barring that, it would be useful to be able to at least point to the directory containing protos so that these could be copied over and referred to explicitly.

kmimu:
This PR addresses the issue mentioned in Issue #404 by implementing [the suggested Option 2](https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/issues/404#issuecomment-1669892639)

- Refactored TraceExporter client initialization
- Replaced direct proto file references with JSON files
- Converted proto files to JSON using the `npm run compile-protos` command

Checks:

Fixes #404


MichaelDoyle:
@aabmass would you be able to take a look at this?

packages/opentelemetry-cloud-trace-exporter/package.json
```diff
@@ -10,11 +10,12 @@
...
     "clean": "gts clean",
     "precompile": "version-update.js",
-    "compile": "tsc",
+    "compile": "tsc && cp -r protos build",
     "prepack": "npm run compile",
     "test": "nyc ts-mocha -p tsconfig.json 'test/**/*.test.ts'",
     "fix": "gts fix",
-    "pretest": "npm run compile"
+    "pretest": "npm run compile",
+    "compile-protos": "node compile-protos.js"
```
aabmass, line 18: Can you add this to run during `precompile` step?

packages/opentelemetry-cloud-trace-exporter/compile-protos.js
```diff
@@ -0,0 +1,60 @@
...
+    '-o',
+    outputFilePath,
+    tracingProtoPath,
+  ];
+
+  try {
+    await pbjsMain(pbjsArgs);
+    console.log('Proto JSON file generated successfully.');
+  } catch (error) {
+    console.error('Error generating Proto JSON file:', error);
```
aabmass, line 56: I think we would want this script to fail. Can you re-throw the error or just not catch the error?

aabmass (commented):
Thank you for the PR! I left a few comments regarding using typescript and a few questions.

packages/opentelemetry-cloud-trace-exporter/src/trace.ts
```diff
@@ -20,14 +20,14 @@ import {
...
 import * as protoloader from '@grpc/proto-loader';
-import * as protofiles from 'google-proto-files';
 import * as grpc from '@grpc/grpc-js';
 import {GoogleAuth} from 'google-auth-library';
 import {promisify} from 'util';
 import {TraceExporterOptions} from './external-types';
 import {getReadableSpanTransformer} from './transform';
 import {TraceService, NamedSpans} from './types';
 import {VERSION} from './version';
+const protoJson = require('../protos/protos.json');
```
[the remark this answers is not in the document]
kmimu: fixed this in [0fb04c4](https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/pull/741/commits/0fb04c4f32218ee2bc9d9a99e881f7d430f3e2f2)

packages/opentelemetry-cloud-trace-exporter/tsconfig.json
```diff
@@ -2,7 +2,8 @@
   "extends": "gts/tsconfig-google.json",
   "compilerOptions": {
     "rootDir": ".",
-    "outDir": "build"
+    "outDir": "build",
+    "resolveJsonModule": true,
```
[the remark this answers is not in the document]
kmimu: No, users of this library will not necessarily need to turn on resolveJsonModule in their TS config unless they are directly importing JSON files in their own project. If library uses JSON internally and is already compiled, the consumers will interact with the built JavaScript or type definition files.

For example, if you add a line like this (and compile):
```
diff --git a/packages/opentelemetry-cloud-trace-exporter/src/trace.ts b/packages/opentelemetry-cloud-trace-exporter/src/trace.ts
index 2b73a36..21e52ee 100644
--- a/packages/opentelemetry-cloud-trace-exporter/src/trace.ts
+++ b/packages/opentelemetry-cloud-trace-exporter/src/trace.ts
@@ -30,6 +30,9 @@ import {VERSION} from './version';
 import * as protoJson from '../protos/protos.json';
 
+export const value = protoJson.nested.google.nested.devtools.nested.cloudtrace.nested.v2.nested.TraceService.methods;
```

You can see in trace.d.ts that the type information can be retrieved without directly referencing the JSON in the user's code:
```
export declare const value: {
    BatchWriteSpans: {
        requestType: string;
        responseType: string;
        options: {
            "(google.api.http).post": string;
            "(google.api.http).body": string;
            "(google.api.method_signature)": string;
....
```

packages/opentelemetry-cloud-trace-exporter/package.json
kmimu: I tested a few scenarios, and it turns out that if we move the JSON file directly into the build directory and remove `/protos/protos.json`, the reference from `trace.ts` results in an error.

However, during my testing, I realized that we don’t need to run `cp -r protos build` because the JSON file is already included in the build due to the include option in the tsconfig.json. Therefore, I have removed that command since it was redundant. → [1558ee4](https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/pull/741/commits/1558ee4d3b7cff4e214b621dd3e1adccef8585b4)

If you know of any alternative ways to handle this without removing `protos/protos.json`, I’d appreciate your idea!

packages/opentelemetry-cloud-trace-exporter/compile-protos.js
kmimu: fixed this in [cf66489](https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/pull/741/commits/cf664897f7b2b496cd8c2f14810cca861bbb49cb)

kmimu:
@aabmass 
Thanks for your initial review! I’ve addressed all the comments, and I’d really appreciate it if you could take another look when you have a chance.

packages/opentelemetry-cloud-trace-exporter/tsconfig.json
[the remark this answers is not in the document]
aabmass: Very clear, thank you 🙂

packages/opentelemetry-cloud-trace-exporter/package.json
aabmass: I guess we could simply add it to .gitiginore.

However I think we what have here is fine–I doubt the generated code will change often anyway. Thank you for looking into it.

aabmass (approved):

PR #762, opened 2024-09-24
chore(deps): bump express to 4.21.0

aabmass:
Express is used in samples. There is a vulnerability in a transitive dep that this fixes.


samples/instrumentation-quickstart/package.json
```diff
@@ -21,7 +21,6 @@
     "posttest": "npm run lint"
   },
   "devDependencies": {
-    "@types/express": "4.17.21",
```
aabmass, line 24: This was accidentally put here and not needed

dashpole (approved):

PR #791, opened 2025-03-28
chore(deps)!: upgrade to otel v2.0.0

fengmk2:
Thank you for opening a Pull Request! Before submitting your PR, there are a few things you can do to make sure it goes smoothly:
- [x] Make sure to open an issue as a [bug/issue/feature](https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [x] Ensure the tests and linter pass
- [ ] Code coverage does not decrease (if any source code was changed)
- [ ] Appropriate docs were updated (if necessary)

Fixes #790

BREAKING CHANGE: drops compatibility with OpenTelemetry SDK 1.0 which changes some types and may require some updates. For more information on the changes and migrating to 2.x, see https://github.com/open-telemetry/opentelemetry-js/blob/v2.0.1/doc/upgrade-to-2.x.md.

Issue #630, opened 2023-10-04
Update supported versions of node


aabmass:
The packages in this repository support all [officially supported Node.js versions](https://nodejs.org/en/about/releases/).

Current supported node versions are 18, 20, (and soon 21).

- [ ] Update CI to support only newest versions
- [ ] Upgrade dependencies which don't support older versions
- [ ] Make a **major version bump release**. Include other upcoming breaking changes at the same time, e.g.
  - [ ] Upgrade to Typescript 5 https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/pull/607
  - [ ] Get rid of deprecated async [GcpDetector](https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/blob/22ffffcd7ad2b97febf0aff51505c08dfb5b1ffe/packages/opentelemetry-resource-util/src/detector/detector.ts#L162-L164)

I'd like to make one more minor version release before doing this.

aabmass:
I am planning to address this when the GCP client libraries also drop support for older node versions.

PR #801, opened 2025-08-27
chore!: drop support for EOL Node.js versions <18

aabmass:
Unblocks https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/issues/630 and https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/pull/791/

BREAKING CHANGE: removes support for older Node.js versions in `package.json` engines and CI jobs.


psx95 (approved):
LGTM

PR #816, opened 2025-08-29
chore(deps): upgrade @types/node to v18

aabmass:
part of https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/issues/630


samples/otlptraceexport/src/override.d.ts
aabmass, line 1: These files (to make typescript happy) are no longer needed with new node type definitions

PR #817, opened 2025-08-29
feat(resource-util)!: remove deprecated code and features

aabmass:
before the next release where we are already making some breaking changes


packages/opentelemetry-resource-util/src/index.ts
```diff
@@ -17,8 +17,6 @@ import {Resource} from '@opentelemetry/resources';
 import {
   CLOUDPLATFORMVALUES_AWS_EC2,
   CLOUDPLATFORMVALUES_GCP_APP_ENGINE,
-  CLOUDPLATFORMVALUES_GCP_CLOUD_FUNCTIONS,
```
psx95, line 20: Does this mean we are dropping support for Cloud Functions and Cloud Run in JS resource detector ?

psx95 (approved):

PR #911, opened 2026-06-11
fix(resource-util): add otel api peer dependency

### LINKED ISSUE #729: in @google-cloud/opentelemetry-resource-util opentelemetry/api is declared as development only dependency, but is actually used in "production".

N/A

N/A, yarn berry in [pnp mode](https://yarnpkg.com/features/pnp)

### What did you do?

I am using modern yarn PNP mode, which ensures that packages only can see dependencies they declare, this leads to code depending on  @google-cloud/opentelemetry-resource-util  being non-deployable unless package is patched to add missing dependency.

application startup, package usable.

### What did you see instead?
```
  '	ERROR: (gcloud.functions.deploy) OperationError: code=3, message=Function failed on loading user code. This is likely due to a bug in the user code. Error message: Provided module can't be loaded.
  '	Is there a syntax error in your code?
  '	Detailed stack trace: Error: @google-cloud/opentelemetry-resource-util tried to access @opentelemetry/api, but it isn't declared in its dependencies; this makes the require call ambiguous and unsound.
  '	Required package: @opentelemetry/api
  '	Required by: @google-cloud/opentelemetry-resource-util@virtual:eecbd1d0b350489fd8e261420b70a0cc0e338d62ee9794a1ab8cd61f7a8ea4723444d65439bea19a80da7175c4eee58bf7a4032bda1a0a3d1faafd08709adedb#npm:2.2.0 (via /workspace/.yarn/__virtual__/@google-cloud-opentelemetry-resource-util-virtual-0d7cba935b/0/cache/@google-cloud-opentelemetry-resource-util-npm-2.2.0-56dcb4a1b6-f771ebeda0.zip/node_modules/@google-cloud/opentelemetry-resource-util/build/src/detector/)
  '	Require stack:
  '	- /workspace/.yarn/__virtual__/@google-cloud-opentelemetry-resource-util-virtual-0d7cba935b/0/cache/@google-cloud-opentelemetry-resource-util-npm-2.2.0-56dcb4a1b6-f771ebeda0.zip/node_modules/@google-cloud/opentelemetry-resource-util/build/src/detector/gce.js
```

### Additional context

root cause:  package [opentelemetry-resource-util](https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/blob/main/packages/opentelemetry-resource-util/package.json#L49C6-L49C24) declares `"@opentelemetry/api"` as devDependencies, but in reality uses it in production inside [detector/gce.ts](https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/blob/main/packages/opentelemetry-resource-util/src/detector/gce.ts#L20)

For others who may be affected, there is workaround, (but it has to be performed in every workspace that uses @google-cloud/opentelemetry-resource-util so is not very scalable), add following to your `.yarnrc.yml`:
```yml
packageExtensions: {
  # TODO: FIXME: https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/issues/729
  "@google-cloud/opentelemetry-resource-util@*": {
    dependencies: {
      "@opentelemetry/api": "*",
    },
  }
}
```

miachillgood:
## Summary
- add `@opentelemetry/api` to `@google-cloud/opentelemetry-resource-util` peer dependencies
- keep the existing devDependency for local build and test workflows
- align the package with the other OpenTelemetry packages in this monorepo that already declare `@opentelemetry/api` as a peer dependency

## Why
Issue #729 reports that `@google-cloud/opentelemetry-resource-util` imports `@opentelemetry/api` at runtime but does not declare it as a runtime-facing dependency. In strict package managers such as Yarn PnP, that makes the package unusable unless consumers add a local package extension.

Adding `@opentelemetry/api` as a peer dependency makes the contract explicit for consumers and matches the dependency shape already used by sibling packages in this repository.

## Validation
- `npm test --workspace @google-cloud/opentelemetry-resource-util`
- `npm run lint --workspace @google-cloud/opentelemetry-resource-util`

Closes #729

### DIFF
--- a/packages/opentelemetry-resource-util/package.json
+++ b/packages/opentelemetry-resource-util/package.json
@@ -62,7 +62,7 @@
     "typescript": "5.0.4"
   },
   "peerDependencies": {
-    "@opentelemetry/api": "^1.9.0",
+    "@opentelemetry/api": "^1.0.0",
     "@opentelemetry/core": "^2.0.0",
     "@opentelemetry/resources": "^2.0.0"
   },
--- a/packages/opentelemetry-resource-util/package.json
+++ b/packages/opentelemetry-resource-util/package.json
@@ -62,6 +62,7 @@
     "typescript": "5.0.4"
   },
   "peerDependencies": {
+    "@opentelemetry/api": "^1.9.0",
     "@opentelemetry/core": "^2.0.0",
     "@opentelemetry/resources": "^2.0.0"
   },


miachillgood:
Updated this branch onto the current `main`.

It still keeps the peer dependency at `^1.0.0` as suggested in review, so the package-side change here should now reflect that feedback plus the latest base branch.

From what I can see, the remaining blocked checks are the usual external e2e gates rather than a code-side issue in this PR. If you want anything else adjusted in the package before the next pass, I can update it.

aabmass (approved):

PR #912, opened 2026-06-11
fix(exporters): respect explicit projectId

### LINKED ISSUE #794: TraceExporter ignores explicitly set GCP projectId, overrides with user's default quota project

22.12.0

### What did you do?
TraceExporter provides the [option](https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/blob/main/packages/opentelemetry-cloud-trace-exporter/src/external-types.ts#L21) to set an explicit GCP projectId for CloudTrace export. 

Based on the API description, the project defined here should be used when explicitly set, or inferred from auth credentials otherwise. In reality however, it appears to always prioritize the default quota-project in the user's auth credentials.

**Repro:**

In your app's instrumentation code:

```
import { TraceExporter } from "@google-cloud/opentelemetry-cloud-trace-exporter";
import { diag, DiagConsoleLogger, DiagLogLevel } from "@opentelemetry/api";

...

const traceExporter = new TraceExporter({
	projectId: [YOUR-PROJECT],
});

...
// log to console if desired
diag.setLogger(new DiagConsoleLogger(), { logLevel: DiagLogLevel.ALL });
```

In a separate terminal, via gcloud cli:

```
gcloud auth application-default set-quota-project [SOME-OTHER-PROJECT]
```

It should use the explicitly set GCP project

```
Google Cloud Trace export
extensionHostProcess.js:179
Google Cloud Trace batch writing traces
extensionHostProcess.js:179
batchWriteSpans successfully
```

### What did you see instead?
It attempts to write traces to the user's default quota project.

```
Google Cloud Trace export
extensionHostProcess.js:179
Google Cloud Trace batch writing traces
extensionHostProcess.js:179
batchWriteSpans error: 7 PERMISSION_DENIED: Cloud Trace API has not been used in project [OTHER-PROJECT] before or it is disabled. Enable it by visiting https://console.developers.google.com/apis/api/cloudtrace.googleapis.com/overview?project=shared-workstations-8s61ch then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry.
```

### Additional context

A workaround in dev environments is to set the env var, but this is not always possible in production.

```
"GOOGLE_CLOUD_QUOTA_PROJECT": [YOUR-PROJECT],
```

miachillgood:
## Summary

- preserve an explicitly configured `projectId` in both the trace and monitoring exporters
- fall back to `GoogleAuth.getProjectId()` only when the caller did not provide `projectId`
- cover the explicit-project and lookup-failure paths for both exporters

## Why

Both exporter constructors pass `projectId` to `GoogleAuth`, but then unconditionally replace the exporter field with the result of `getProjectId()`. Retaining the explicit value prevents ambient project discovery from changing the destination project selected by the exporter.

This does not change quota-project handling. The quota project remains credential-driven and is used separately for the `x-goog-user-project` header; this change only controls the project ID used by the exporters for destination resource names.

## Validation

- `npm test --workspace @google-cloud/opentelemetry-cloud-trace-exporter` (30 passing)
- `npm run lint --workspace @google-cloud/opentelemetry-cloud-trace-exporter`
- `npm test --workspace @google-cloud/opentelemetry-cloud-monitoring-exporter` (39 passing)
- `npm run lint --workspace @google-cloud/opentelemetry-cloud-monitoring-exporter`

The trace lint command reports two pre-existing warnings in `test/transform.test.ts`; there are no lint errors.

Closes #794

### DIFF
--- a/packages/opentelemetry-cloud-trace-exporter/src/trace.ts
+++ b/packages/opentelemetry-cloud-trace-exporter/src/trace.ts
@@ -67,11 +67,15 @@ export class TraceExporter implements SpanExporter {
       scopes: ['https://www.googleapis.com/auth/cloud-platform'],
     });
 
-    // Start this async process as early as possible. It will be
-    // awaited on the first export because constructors are synchronous
-    this._projectId = this._auth.getProjectId().catch(err => {
-      diag.error(err);
-    });
+    if (options.projectId) {
+      this._projectId = options.projectId;
+    } else {
+      // Start this async process as early as possible. It will be
+      // awaited on the first export because constructors are synchronous
+      this._projectId = this._auth.getProjectId().catch(err => {
+        diag.error(err);
+      });
+    }
 
     if (options.apiEndpoint) {
       this._apiEndpoint = options.apiEndpoint;
--- a/packages/opentelemetry-cloud-monitoring-exporter/src/monitoring.ts
+++ b/packages/opentelemetry-cloud-monitoring-exporter/src/monitoring.ts
@@ -105,11 +105,15 @@ export class MetricExporter implements PushMetricExporter {
       ),
     });
 
-    // Start this async process as early as possible. It will be
-    // awaited on the first export because constructors are synchronous
-    this._projectId = this._auth.getProjectId().catch(err => {
-      diag.error(err);
-    });
+    if (options.projectId) {
+      this._projectId = options.projectId;
+    } else {
+      // Start this async process as early as possible. It will be
+      // awaited on the first export because constructors are synchronous
+      this._projectId = this._auth.getProjectId().catch(err => {
+        diag.error(err);
+      });
+    }
   }
 
   /**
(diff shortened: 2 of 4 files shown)


packages/opentelemetry-cloud-trace-exporter/src/trace.ts
```diff
@@ -165,8 +169,26 @@ export class TraceExporter implements SpanExporter {
...
+              (headers as {entries: () => Iterable<[string, string]>}).entries()
+            )
+          );
+        }
+        return Object.fromEntries(
+          Object.entries(headers as Record<string, string>).map(([key, value]) => [
+            key.toLowerCase(),
+            value,
+          ])
+        );
```
aabmass: Why was this change needed? The code was fully typed before

aabmass:
Does this change have any impact on the quota project?

packages/opentelemetry-cloud-trace-exporter/src/trace.ts
aabmass: I looked at the new push, still not sure why we need to lower case either

miachillgood:
Thanks for checking. This does not change quota-project selection. It only preserves the explicitly configured destination project used by the exporters. The quota project remains credential-driven and continues to control the `x-goog-user-project` header.

I also removed the unrelated header normalization, rebuilt the branch on the current `main`, and added coverage for both explicit configuration and lookup failure in the trace and monitoring exporters.

PR #932, opened 2026-07-31
fix(deps): update googleapis and gcp-metadata to fix npm audit warnings

aabmass:
Fixes all prod security issues from `npm audit` in the production npm packages (not necessarily all samples).

- Upgrades `googleapis` (`^173.0.0`), `gcp-metadata` (`^9.0.0`), and `google-auth-library` (`^10.0.0`) to drop vulnerable `uuid < 11.1.1` from production dependencies.
- Fixes `@grpc/grpc-js` and `google-auth-library` v10 WHATWG `Headers` incompatibility (grpc/grpc-node#2993, googleapis/google-auth-library#1798) by adapting `creds.getRequestHeaders()` to return plain key-value objects in `trace-exporter` and samples.
- Shims global `File` type in `dom-shim.d.ts` for `gaxios` (googleapis/google-cloud-node#8107).

### DIFF
--- a/samples/instrumentation-quickstart/src/instrumentation.ts
+++ b/samples/instrumentation-quickstart/src/instrumentation.ts
@@ -72,7 +72,7 @@ function getMetricReader() {
       return undefined;
     default:
       throw Error(
-        `no valid option for OTEL_METRICS_EXPORTER: ${process.env.OTEL_METRICS_EXPORTER}`
+        `no valid option for OTEL_METRICS_EXPORTER: ${process.env.OTEL_METRICS_EXPORTER}`,
       );
   }
 }
@@ -82,8 +82,8 @@ function getMetricReader() {
 diag.setLogger(
   new DiagConsoleLogger(),
   opentelemetry.core.diagLogLevelFromString(
-    opentelemetry.core.getStringFromEnv('OTEL_LOG_LEVEL')
-  )
+    opentelemetry.core.getStringFromEnv('OTEL_LOG_LEVEL'),
+  ),
 );
 
 const sdk = new opentelemetry.NodeSDK({
@@ -101,7 +101,7 @@ try {
 } catch (error) {
   diag.error(
     'Error initializing OpenTelemetry SDK. Your application is not instrumented and will not produce telemetry',
-    error
+    error,
   );
 }
 
--- a/samples/otlpmetricexport/package.json
+++ b/samples/otlpmetricexport/package.json
@@ -24,8 +24,8 @@
     "@opentelemetry/resources": "2.0.1",
     "@opentelemetry/sdk-node": "0.217.0",
     "@opentelemetry/sdk-metrics": "2.0.1",
-    "@grpc/grpc-js": "1.14.3",
-    "google-auth-library": "9.15.1"
+    "@grpc/grpc-js": "1.14.4",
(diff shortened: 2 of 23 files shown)


samples/otlpmetricexport/package.json
```diff
@@ -24,8 +24,8 @@
     "@opentelemetry/resources": "2.0.1",
     "@opentelemetry/sdk-node": "0.217.0",
     "@opentelemetry/sdk-metrics": "2.0.1",
-    "@grpc/grpc-js": "1.14.3",
-    "google-auth-library": "9.15.1"
+    "@grpc/grpc-js": "^1.14.3",
```
aabmass: Why did you add a caraet???

samples/otlptraceexport/package.json
```diff
@@ -21,16 +21,16 @@
...
     "@opentelemetry/api": "1.9.0",
     "@opentelemetry/exporter-trace-otlp-grpc": "0.203.0",
     "@opentelemetry/exporter-trace-otlp-http": "0.203.0",
     "@opentelemetry/exporter-trace-otlp-proto": "0.203.0",
     "@opentelemetry/resources": "2.0.1",
     "@opentelemetry/sdk-node": "0.217.0",
     "axios": "1.15.2",
-    "express": "4.22.1",
-    "google-auth-library": "9.15.1"
+    "express": "^4.22.1",
```
aabmass: same question. this is a sample (not a library) so it's ok to use pinned deps.

samples/instrumentation-quickstart/src/instrumentation.ts
aabmass, line 1: I see a bunch of seemingly unneded diffs like this (a few other files too). Audit the diff of this PR to check if you can revert some of these changes to reduce noise.

samples/otlptraceexport/src/app-http-proto-export.ts
```diff
@@ -32,7 +32,8 @@ async function getAuthenticatedClient(): Promise<AuthClient> {
 // Express App that exports traces over HTTP with protobuf
 async function main() {
   const authenticatedClient = await getAuthenticatedClient();
-  const requestHeaders = await authenticatedClient.getRequestHeaders();
+  const rawHeaders = await authenticatedClient.getRequestHeaders();
+  const requestHeaders = Object.fromEntries(rawHeaders.entries());
```
psx95, line 36: Question: What prompted this?

psx95 (approved):

samples/otlptraceexport/src/app-http-proto-export.ts
aabmass: It's similar to It's this one https://github.com/grpc/grpc-node/issues/2993. The [OTLP exporter expects a map string -> string](https://open-telemetry.github.io/opentelemetry-js/interfaces/_opentelemetry_otlp-exporter-base.index.OTLPExporterNodeConfigBase.html#headers) and not this native `Headers` class

But now I'm noticing the exporter also accepts now accepts a factory function, will update.

aabmass: > But now I'm noticing the exporter also accepts now accepts a factory function, will update.

hmm actually that's not supported in the version of the OTLP exporters used in this PR. I'll send a follow up.

Issue #781, opened 2024-12-13
Add notice that x-cloud-trace-header is deprecated/legacy


hongalex:
Google's `x-cloud-trace-header` is legacy/deprecated per https://cloud.google.com/trace/docs/trace-context#legacy-http-header

The usage of it [here](https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/blob/f8b515719c27010c1b4c6f7a34e6a4fc512a9c0b/packages/opentelemetry-cloud-trace-propagator/src/CloudPropagator.ts#L30) should be discouraged / marked deprecated

aabmass:
Thanks for filing, we should probably do this to make it clear to users.

PR #933, opened 2026-08-03
docs: mark cloud trace header as legacy

miachillgood:
## Summary
- mark `x-cloud-trace-context` as a legacy Google Cloud Trace header in the propagator source comment
- add the same guidance to the package README and recommend W3C Trace Context where possible

## Why
Addresses #781 by making it clearer that this header format is legacy/deprecated and should mainly be used for compatibility scenarios.

## Testing
- not run (docs-only change)


aabmass (approved):
Thanks!

psx95 (approved):

PR #936, opened 2026-08-03
docs: update otlp samples to use header provider

aabmass:
So that auth refresh happens. This is a new-ish feature for the OTLP exporters from https://github.com/open-telemetry/opentelemetry-js/pull/5994


psx95 (commented):
I'd argue to make this change directly in https://github.com/GoogleCloudPlatform/opentelemetry-samples/tree/main/javascript/otlptraceexport/src

aabmass:
Thanks, should we delete these samples altogether ?

psx95:
> Thanks, should we delete these samples altogether ?

You can, but I just left them as-is after marking the repo as deprecated in other languages.

PR #946, opened 2026-08-04
docs: add deprecation notices and migration guide in favor of native OTLP endpoints

aabmass:
This change marks all packages in the repository as deprecated and directs users to migrate to standard OpenTelemetry OTLP exporters and the Cloud Telemetry API.

### Changes
* Adds `MIGRATION.md` at the repository root with step-by-step instructions, mapping tables, and TypeScript examples for Trace, Metrics, and Trace Propagator migration (inspired by the reference [Java Migration Guide](https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/blob/main/MIGRATION.md) and [Python Migration Guide](https://github.com/GoogleCloudPlatform/opentelemetry-operations-python/blob/main/MIGRATION.md)).
* Adds deprecation notices to the root `README.md` and all individual package READMEs (`opentelemetry-cloud-trace-exporter`, `opentelemetry-cloud-monitoring-exporter`, `opentelemetry-cloud-trace-propagator`, `opentelemetry-resource-util`).

### Follow-up Steps
- Updating the code with deprecation notices #948 
- After these changes are merged and a final release is published, we will run `npm deprecate` across all 4 published npm packages so that `npm install` displays a CLI deprecation warning pointing developers to the migration guide.

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -71,25 +71,7 @@ npm install @opentelemetry/exporter-trace-otlp-proto # Recommended for HTTP/Prot
 
 ### 2. Configure the SDK
 
-Standard OTel JS environment variables automatically append protocol path suffixes (`/v1/traces`), but passing a URL explicitly to the programmatic constructor does **not**. When exporting telemetry data to Google Cloud, you can configure your exporter using one of the following supported patterns:
-
-**Option A: Environment Variables (Recommended)**
-Do not pass arguments to the constructor; it will automatically honor environment variables and append protocol paths. This option is ideal when authentication is handled externally (for example, when exporting to an OpenTelemetry Collector):
-
-```bash
-export OTEL_EXPORTER_OTLP_ENDPOINT="https://telemetry.googleapis.com"
-export OTEL_TRACES_EXPORTER="otlp"
-export OTEL_EXPORTER_OTLP_PROTOCOL="http/protobuf"
-```
-
-```typescript
-import { OTLPTraceExporter } from '@opentelemetry/exporter-trace-otlp-proto';
-
-const exporter = new OTLPTraceExporter(); // Automatically points to https://telemetry.googleapis.com/v1/traces
-```
-
-**Option B: Programmatic Configuration with Dynamic Google Auth (TypeScript)**
-You can export OTLP telemetry directly from your application to Google Cloud endpoints (`https://telemetry.googleapis.com`) without external infrastructure by using `google-auth-library` to supply Application Default Credentials (ADC). Because OAuth2 tokens expire after 1 hour, standard OpenTelemetry JS OTLP exporters natively support async header callbacks (for HTTP) and channel credentials wrapping (for gRPC) to dynamically supply fresh tokens:
+When exporting OTLP telemetry directly from your application to Google Cloud endpoints (`https://telemetry.googleapis.com`), you can use `google-auth-library` to supply Application Default Credentials (ADC). Because OAuth2 tokens expire after 1 hour, standard OpenTelemetry JS OTLP exporters natively support async header callbacks (for HTTP) and channel credentials wrapping (for gRPC) to dynamically supply fresh tokens:
 
 ```bash
 npm install google-auth-library
@@ -164,11 +146,6 @@ async function main(): Promise<void> {
 
 main().catch(console.error);
 ```
-
-**Option C: Export via OpenTelemetry Collector**
-Teams wishing to offload token lifecycle management, batching, and retries from application code can send telemetry to a local **OpenTelemetry Collector** (e.g., `http://localhost:4318`) running as a sidecar or gateway, configured with the `googleclientauth` extension and `otlphttp` / `googlecloud` exporter.
-
-*For complete end-to-end TypeScript implementations, see the official [GoogleCloudPlatform/opentelemetry-samples](https://github.com/GoogleCloudPlatform/opentelemetry-samples/tree/main/javascript) repository.*
 
 ### 3. Follow the Migration Guide
 
@@ -295,12 +272,12 @@ counter.add(1, { job_type: 'import', status: 'success' });
 Run both the legacy exporter and the OTLP exporter concurrently:
 
 ```typescript
-import { MetricExporter as LegacyMetricExporter } from '@google-cloud/opentelemetry-cloud-monitoring-exporter';
+import { MetricExporter } from '@google-cloud/opentelemetry-cloud-monitoring-exporter';
 import { OTLPMetricExporter } from '@opentelemetry/exporter-metrics-otlp-proto';
 import { NodeSDK } from '@opentelemetry/sdk-node';
 import { PeriodicExportingMetricReader } from '@opentelemetry/sdk-metrics';
 
-const legacyExporter = new LegacyMetricExporter();
+const legacyExporter = new MetricExporter();
 const otlpExporter = new OTLPMetricExporter({
   url: 'https://telemetry.googleapis.com/v1/metrics',
 });
@@ -318,7 +295,7 @@ sdk.start();
 
 1. **Verify New Metrics Ingestion:** Once double-writing is deployed, verify in Metrics Explorer that new metrics are arriving successfully under the `prometheus.googleapis.com/` domain.
 2. **Update Dashboards & Alerts:** Duplicate or update existing Cloud Monitoring dashboards and alerting policies to query `prometheus.googleapis.com/` metric names instead of `workload.googleapis.com/`.
-3. **Decommission:** Once all dashboards are verified against the new OTLP metric data streams, remove `LegacyMetricExporter` to complete the migration and eliminate double-ingestion costs.
+3. **Decommission:** Once all dashboards are verified against the new OTLP metric data streams, remove `MetricExporter` to complete the migration and eliminate double-ingestion costs.
 
 ---
 
@@ -432,19 +409,9 @@ The following features of the legacy `MetricExporter` are not supported in stand
 * **Custom Prefix via Exporter Option (`prefix`)**: Legacy exporter accepted a `prefix` constructor option. Standard OTLP exporters ingest under `prometheus.googleapis.com/` by default. Preserving prefixes requires instrumentation changes or Strategy 3 (wrapped exporter).
 * **Custom User-Agent Overrides (`userAgent`)**: Custom User-Agent overrides via exporter options are not supported in standard OTLP exporters; use standard OpenTelemetry resource attributes (`service.name`, `service.version`).
 
-### Telemetry API Server-Side Conversion Nuances
+The Google Cloud Telemetry API converts OTLP metric data according to the official [Google Cloud Telemetry API Metric Mapping specification](https://docs.cloud.google.com/stackdriver/docs/reference/telemetry/v1.metrics#metric-mapping-reference-info).
 
-The backend converts OTLP metric data according to the [Google Cloud Telemetry API Metric Mapping specification](https://docs.cloud.google.com/stackdriver/docs/reference/telemetry/v1.metrics#metric-mapping-reference-info). Key differences include:
-
-* **Value Types (INT64 vs DOUBLE):** The Telemetry API translates **all OTLP `INT64` scalar metrics to `DOUBLE`** in Cloud Monitoring to prevent Monarch value-type schema collisions.
-* **Metric Kind & Temporality:** 
-  - Monotonic cumulative sums map to `CUMULATIVE` (suffixed with `/counter`).
-  - Monotonic delta sums map to `DELTA` (suffixed with `/delta`).
-  - Non-monotonic sums map to `GAUGE` (suffixed with `/gauge`). Non-monotonic delta sums are not supported.
-  - **Histograms:** Support both `CUMULATIVE` (suffixed with `/histogram`) and `DELTA` (suffixed with `/histogram:delta`) distributions.
-  - **Summary Metrics:** Expanded into individual time series for `_count` (`CUMULATIVE`), `_sum` (`CUMULATIVE`), and `quantile` (`GAUGE` with a `quantile` label).
-* **Resource Attributes & `target_info` Metric:** The API maps resources server-side and automatically generates a `target_info` metric for each unique OpenTelemetry resource containing non-identifying resource attributes.
-* **Special Characters:** The Telemetry API preserves `.` and `/` characters in OTLP metric names rather than replacing them with underscores (`_`).
+Please refer to the specification documentation for complete, up-to-date details on how metric kinds, temporality, value types (`INT64` to `DOUBLE`), special characters, and resource attributes are mapped server-side in Google Cloud Monitoring.
 
 ### Mapping and Limitations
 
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -1,3 +1,8 @@
+> [!WARNING]
+> This repository and all its packages are **deprecated**. Google Cloud Platform now supports native OpenTelemetry Protocol (OTLP) endpoints via the [Telemetry API](https://cloud.google.com/stackdriver/docs/reference/telemetry/overview). Please migrate to use community-maintained OTLP exporters instead.
+> 
+> See the [Migration Guide](https://docs.cloud.google.com/stackdriver/docs/instrumentation/migrate-to-otlp-endpoints) for traces, and the [Telemetry API Documentation](https://cloud.google.com/stackdriver/docs/reference/telemetry/overview) for metrics.
```
psx95: We should link to the Migration guide added in this repository. Once the migration instructions are available in the public GCP docs, we can update the contents in the Migration Guide to simply point to the correct URLs.

packages/opentelemetry-cloud-trace-exporter/src/trace.ts
psx95, line 1: Not sure if it is the idiomatic way in JS, but for good measure, we should also put the deprecation notice in a log that is [emitted once the exporters are initialized (*we did this for Java*)](https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/blob/59abf06f16b2e61becd6a17999371cf883d91a7f/exporters/trace/src/main/java/com/google/cloud/opentelemetry/trace/TraceExporter.java#L35) along with the lint annotations.

MIGRATION.md
```diff
@@ -0,0 +1,486 @@
...
+      ),
+    }),
+  });
+  sdk.start();
+}
+
+main().catch(console.error);
+```
+
+**Option C: Export via OpenTelemetry Collector**
```
psx95: IMO this does not belong in a migration guide as this option is somewhat orthogonal.

MIGRATION.md
```diff
@@ -0,0 +1,486 @@
...
+npm uninstall @google-cloud/opentelemetry-cloud-trace-exporter
+npm install @opentelemetry/exporter-trace-otlp-proto # Recommended for HTTP/Protobuf
+# or npm install @opentelemetry/exporter-trace-otlp-http (for HTTP/JSON)
+```
+
+### 2. Configure the SDK
+
+Standard OTel JS environment variables automatically append protocol path suffixes (`/v1/traces`), but passing a URL explicitly to the programmatic constructor does **not**. When exporting telemetry data to Google Cloud, you can configure your exporter using one of the following supported patterns:
+
+**Option A: Environment Variables (Recommended)**
```
psx95: QQ - How does auth work here?

MIGRATION.md
```diff
@@ -0,0 +1,486 @@
...
+      },
+    }),
+  });
+  sdk.start();
+}
+
+main().catch(console.error);
+```
+
+*OTLP/gRPC Dynamic Auth Example:*
```
psx95: Could we link to the [app-grpc-export sample](https://github.com/GoogleCloudPlatform/opentelemetry-samples/blob/main/javascript/otlptraceexport/src/app-grpc-export.ts) here instead of writing a new one?

This looks similar.

MIGRATION.md
```diff
@@ -0,0 +1,486 @@
...
+
+Run both the legacy exporter and the OTLP exporter concurrently:
+
+```typescript
+import { MetricExporter as LegacyMetricExporter } from '@google-cloud/opentelemetry-cloud-monitoring-exporter';
+import { OTLPMetricExporter } from '@opentelemetry/exporter-metrics-otlp-proto';
+import { NodeSDK } from '@opentelemetry/sdk-node';
+import { PeriodicExportingMetricReader } from '@opentelemetry/sdk-metrics';
+
+const legacyExporter = new LegacyMetricExporter();
```
psx95: I think the class name is wrong here - should be `MetricExporter`.

MIGRATION.md
```diff
@@ -0,0 +1,486 @@
...
+
+```typescript
+import { propagation } from '@opentelemetry/api';
+import { W3CTraceContextPropagator } from '@opentelemetry/core';
+
+// Set the global propagator to use standard W3C Trace Context
+propagation.setGlobalPropagator(new W3CTraceContextPropagator());
+```
+
+> [!NOTE]
```
psx95: nit: Should this note be just below the "Migrate from Google Cloud Trace Propagator" heading?

MIGRATION.md
```diff
@@ -0,0 +1,486 @@
...
+```
+
+### Unsupported Features
+
+The following features of the legacy `MetricExporter` are not supported in standard OTLP exporters:
+* **Metric Descriptor Strategy (`disableCreateMetricDescriptors`)**: Legacy `MetricExporter` allowed disabling metric descriptor creation. In OTLP, metadata is handled automatically by the backend Telemetry API; client-side descriptor toggling is inapplicable.
+* **Custom Prefix via Exporter Option (`prefix`)**: Legacy exporter accepted a `prefix` constructor option. Standard OTLP exporters ingest under `prometheus.googleapis.com/` by default. Preserving prefixes requires instrumentation changes or Strategy 3 (wrapped exporter).
+* **Custom User-Agent Overrides (`userAgent`)**: Custom User-Agent overrides via exporter options are not supported in standard OTLP exporters; use standard OpenTelemetry resource attributes (`service.name`, `service.version`).
+
+### Telemetry API Server-Side Conversion Nuances
```
psx95: I wonder if this will cause more confusion. Should we just link to the GCP docs instead of maintaining a list of nuances here? 
The GCP docs would have the most up-to-date information on mapping (*even though this is unlikely to change*).

MIGRATION.md
aabmass: It doesn't, sorry. Gemini went off the rails here.

MIGRATION.md
aabmass: I added it inline because it's in Java and Python migration guides too e.g. https://github.com/GoogleCloudPlatform/opentelemetry-operations-python/blob/main/MIGRATION.md#3-programmatic-configuration

What do you prefer?

MIGRATION.md
aabmass: Sorry Gemini went to wild.

README.md
aabmass: Oops forgot to update this one back, thanks.

packages/opentelemetry-cloud-trace-exporter/src/trace.ts
aabmass: Thanks. I split the code changes into a separate PR https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/pull/948

MIGRATION.md
aabmass: Good catch, fixed.

MIGRATION.md
aabmass: Done, moved it up under the heading.

MIGRATION.md
aabmass: Agreed, replaced the list with a link to the mapping spec.

MIGRATION.md
psx95: I think this can be language dependent. I [didn't include it in Java Migration guide](https://github.com/GoogleCloudPlatform/opentelemetry-operations-java/blob/main/MIGRATION.md#strategy-1-direct-migration-recommended) due to the autoconfigure module.

But IIRC in python this was necessary to show. I'm ok with this being here. However, we could just show one (e.g., gRPC), and for the other we could say "See the http example here".

aabmass: Done

packages/opentelemetry-cloud-trace-propagator/README.md
psx95, line 1: QQ - is this propagator published as an artifact and are users expected to use this directly ?

MIGRATION.md
```diff
@@ -0,0 +1,447 @@
...
+
+### Mapping and Limitations
+
+#### Configuration Mapping
+
+| `TraceExporter` Parameter | OTLP Equivalent Property / Env Var | Notes |
+| :--- | :--- | :--- |
+| `projectId` | Resource attribute: `gcp.project_id` | Set via `OTEL_RESOURCE_ATTRIBUTES="gcp.project_id=your-project-id"` or detected automatically via `@opentelemetry/resource-detector-gcp`. |
+| `credentials` / `keyFilename` / `keyFile` | OTLP Exporter Configuration | Handled dynamically via `google-auth-library` or externally via the OpenTelemetry Collector. |
+| `apiEndpoint` | `otel.exporter.otlp.endpoint` / `OTEL_EXPORTER_OTLP_ENDPOINT` | Set it to `https://telemetry.googleapis.com`. |
```
dashpole: Looks like `otel.exporter.otlp.endpoint` is a Java SDK system property rather than a JS property. In JS, this is configured via the `url` option or the `OTEL_EXPORTER_OTLP_ENDPOINT` environment variable:
```suggestion
| `apiEndpoint` | `url` / `OTEL_EXPORTER_OTLP_ENDPOINT` | Set it to `https://telemetry.googleapis.com`. |
```

MIGRATION.md
```diff
@@ -0,0 +1,447 @@
...
+### Mapping and Limitations
+
+#### Configuration Mapping
+
+| `MetricExporter` Parameter | OTLP Equivalent Property / Env Var | Notes |
+| :--- | :--- | :--- |
+| `projectId` | Resource attribute: `gcp.project_id` | Set via `OTEL_RESOURCE_ATTRIBUTES` or detected automatically. |
+| `credentials` / `keyFilename` / `keyFile` | OTLP Exporter Configuration | Handled dynamically via `google-auth-library` or externally via the OpenTelemetry Collector. |
+| `prefix` (default `workload.googleapis.com`) | N/A | Legacy ingested under `workload.googleapis.com/`. OTLP ingests under `prometheus.googleapis.com/` by default. See Strategy 3 for preserving prefixes. |
+| `apiEndpoint` | `otel.exporter.otlp.endpoint` / `OTEL_EXPORTER_OTLP_ENDPOINT` | Set it to `https://telemetry.googleapis.com`. |
```
dashpole: Same here, `otel.exporter.otlp.endpoint` looks like a leftover Java system property:
```suggestion
| `apiEndpoint` | `url` / `OTEL_EXPORTER_OTLP_ENDPOINT` | Set it to `https://telemetry.googleapis.com`. |
```

MIGRATION.md
```diff
@@ -0,0 +1,447 @@
...
+
+This guide provides instructions on how to migrate from the custom exporters and propagators in this repository to the standard OpenTelemetry OTLP exporters and standard W3C Trace Context propagation.
+
+## Overview
+
+Google Cloud supports native OTLP (OpenTelemetry Protocol) ingestion for Cloud Trace and Cloud Monitoring via the [Telemetry API](https://docs.cloud.google.com/stackdriver/docs/reference/telemetry/overview). This allows you to use standard OpenTelemetry OTLP exporters for sending telemetry data to Google Cloud.
+
+## Deprecation Notice
+
+All packages in this repository (`@google-cloud/opentelemetry-cloud-trace-exporter`, `@google-cloud/opentelemetry-cloud-monitoring-exporter`, `@google-cloud/opentelemetry-cloud-trace-propagator`, and `@google-cloud/opentelemetry-resource-util`) are deprecated. **They will be archived after September 30th, 2026.** Please migrate to standard OTLP exporters and standard OpenTelemetry libraries before this date.
```
dashpole: Do we need to bump this back further?  This is less than 2 months away. I would expect a timeline closer to 3-6 months?

MIGRATION.md
psx95: Missed that, good catch David!

MIGRATION.md
psx95: We typically target 90 days, the date was true for Java, but yes we could move to October 30th, 2026 (Technically after September 30th).

aabmass: I'll update to Octoboer.

MIGRATION.md
aabmass: My Jetski missed it :(

packages/opentelemetry-cloud-trace-propagator/README.md
aabmass: Yes check the sample below. But we don't recommend this really anywhere anymore. There's also https://github.com/GoogleCloudPlatform/opentelemetry-operations-js/pull/933 which I will merge right now.

dashpole:
LGTM, although I didn't review the JS code very closely.

MIGRATION.md
```diff
@@ -0,0 +1,447 @@
...
+      },
+    }),
+  });
+  sdk.start();
+}
+
+main().catch(console.error);
+```
+
+*For gRPC export using Application Default Credentials, see the official [`app-grpc-export.ts`](https://github.com/GoogleCloudPlatform/opentelemetry-samples/blob/main/javascript/otlptraceexport/src/app-grpc-export.ts) sample in the `opentelemetry-samples` repository.*
```
aabmass: Should link both samples even thought the http one is inlined too.

MIGRATION.md
```diff
@@ -0,0 +1,447 @@
...
+
+```typescript
+import { OTLPMetricExporter } from '@opentelemetry/exporter-metrics-otlp-proto';
+import { NodeSDK } from '@opentelemetry/sdk-node';
+import { PeriodicExportingMetricReader } from '@opentelemetry/sdk-metrics';
+
+const exporter = new OTLPMetricExporter({
+  url: 'https://telemetry.googleapis.com/v1/metrics',
+  // Note: For direct in-app export to GCP, include the async headers() callback
+  // or gRPC credentials shown in the Trace Exporter section above.
```
aabmass: Let's just include it for copy pasting since this is the recommended approach. We should also add a link to the samples in opentelemetry-samples repo like we did for the trace exporter

MIGRATION.md
```diff
@@ -0,0 +1,447 @@
...
+
+```typescript
+import { metrics } from '@opentelemetry/api';
+import { MeterProvider, PeriodicExportingMetricReader } from '@opentelemetry/sdk-metrics';
+import { MetricExporter } from '@google-cloud/opentelemetry-cloud-monitoring-exporter';
+import { OTLPMetricExporter } from '@opentelemetry/exporter-metrics-otlp-proto';
+
+const legacyExporter = new MetricExporter();
+const otlpExporter = new OTLPMetricExporter({
+  url: 'https://telemetry.googleapis.com/v1/metrics',
```
aabmass:
```suggestion
  url: 'https://telemetry.googleapis.com/v1/metrics',
  // ...
```

MIGRATION.md
```diff
@@ -0,0 +1,447 @@
...
+      exportIntervalMillis: 60000,
+    }),
+    new PeriodicExportingMetricReader({
+      exporter: otlpExporter,
+      exportIntervalMillis: 60000,
+    }),
+  ],
+});
+
+metrics.setGlobalMeterProvider(meterProvider);
```
aabmass: why not use the same set up as above with NodeSDK?

MIGRATION.md
```diff
@@ -0,0 +1,453 @@
...
+```bash
+npm uninstall @google-cloud/opentelemetry-cloud-monitoring-exporter
+npm install @opentelemetry/exporter-metrics-otlp-proto @opentelemetry/resource-detector-gcp google-auth-library
+# If using gRPC:
+# npm install @opentelemetry/exporter-metrics-otlp-grpc @grpc/grpc-js
+# If using HTTP/JSON:
+# npm install @opentelemetry/exporter-metrics-otlp-http
+```
+
+#### 2. Configure the SDK
```
psx95, line 210: The `gcp.project_id` resource attribute would be required here as well, correct?

MIGRATION.md
```diff
@@ -0,0 +1,453 @@
...
+
+### 2. Configure the SDK
+
+Replace `CloudPropagator` with `W3CTraceContextPropagator`:
+
+```typescript
+import { propagation } from '@opentelemetry/api';
+import { W3CTraceContextPropagator } from '@opentelemetry/core';
+
+// Set the global propagator to use standard W3C Trace Context
```
psx95: nit/suggestion: Reference the note above to mention that this will most likely, not be required ?

psx95 (approved):
A couple of nits/questions, but otherwise LGTM

MIGRATION.md
aabmass: Added envvars. And I updated all of the instructions to use `OTEL_EXPORTER_OTLP_ENDPOINT="https://telemetry.googleapis.com"` instead of passing url manually. Tested it all again too.

